PR #16, opened 2012-10-06
robustness fix to _parse_extinf

deutrino:
function now handles titles with commas, and coerces unset durations to 0


leandromoreira:
@jbochi @feuvan and @gordon-morehouse is it still valid to apply or should we wait for test case?

PR #40, opened 2015-03-03
Replace arrow with iso8601 that is more lightweight (fixes #38)

### LINKED ISSUE #38: Remove arrow depencency
I think this looks like a nice and simple library for parsing m3u8 playlists, but one thing that kind of ruined that impression was when I realized that it depended on arrow. I'm not saying arrow is a bad library, but it's way to big (with all it's own dependencies) for what it's used for in this project.

Currently arrow is used for two operations:
* One in parser.py: `arrow.get(value).datetime`...
* ... and one in models.py: `arrow.get(self.program_date_time).isoformat()`

So it's basically parsing and formatting date and times in ISO 8601 format. A more lightweight library without additional dependencies for parsing ISO 8601 is https://pypi.python.org/pypi/iso8601. The operations currently done with arrow can be done like this with iso8601:
* `iso8601.parse_date(value)`
* `self.program_date_time.isoformat()`

Completely removing parsing (just providing the value as a string) or making it optional is another way to go.

netrounds-fredrik:
This PR fixes #38


flavioribeiro:
thanks! :beers:

PR #41, opened 2015-04-23
Check for uri in segment key before creating class

GreenZapdos:
On some m3u8 servers, there is an #EXT-X-KEY line without a uri to signify that there is not a key.

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -361,7 +361,7 @@ class Segment(BasePathMixin):
         self.byterange = byterange
         self.program_date_time = program_date_time
         self.discontinuity = discontinuity
-        self.key = Key(base_uri=base_uri,**key) if key else None
+        self.key = Key(base_uri=base_uri,**key) if key and 'uri' in key else None
 
 
     def dumps(self, last_segment):


flavioribeiro:
@GreenZapdos can you add tests and an example playlist so we can understand better your pull request?

GreenZapdos:
The current version of m3u8 (0.2.4) will not load a playlist like this:

~~~m3u
#EXTM3U

#EXT-X-KEY:METHOD=AES-128,URI="http://example.com/key1",IV=0x01
#EXTINF:10,
http://example.com/video/encrypted-1.ts

#EXT-X-KEY:METHOD=AES-128,URI="http://example.com/key2",IV=0x02
#EXTINF:10,
http://example.com/video/encrypted-2.ts

#EXT-X-KEY:METHOD=AES-128,URI="http://example.com/key3",IV=0x03
#EXTINF:10,
http://example.com/video/encrypted-3.ts

#EXT-X-KEY:METHOD=UNENCRYPTED
#EXTINF:10,
http://example.com/video/unencrypted-A.ts
~~~

I'm not sure how this should be tested.

leandromoreira:
@GreenZapdos I think you can inspire by looking at https://github.com/globocom/m3u8/blob/master/tests/test_parser.py#L10 but instead of a `SIMPLE_PLAYLIST` you could create another type like `PLAYLIST_WITH_EXTKEY_NO_URI` https://github.com/globocom/m3u8/blob/master/tests/playlists.py#L10 and make sure it will pass only if your PR is applied.

leandromoreira:
@GreenZapdos ping?

PR #48, opened 2015-07-21
Add support for #EXT-X-INDEPENDENT-SEGMENTS

mvmocanu:
We would like to be able to preserve `#EXT-X-INDEPENDENT-SEGMENTS` defined here:
https://tools.ietf.org/html/draft-pantos-http-live-streaming-13#section-3.4.16

So I added support to preserve this property.


leandromoreira:
Thanks @mvmocanu , this seems to fix #51 and fix #52

PR #50, opened 2015-07-22
Added a strict mode (strict=True optional parameter to M3U8 constructor), plus a couple of parse fixes.

GrumpyOldTroll:


leandromoreira:
Thanks @GrumpyOldTroll

Issue #42, opened 2015-05-19
Incorrect child absolute_uri values generated from base_uri.


GrumpyOldTroll:
Passing the original url as the base_uri input parameter to M3U8 constructor (as the description in its help text describes) makes incorrect absolute_uri paths for the child objects.

This code exercises the problem against one of Apple's sample streams:
```
$ python
Python 2.7.6 (default, Sep  9 2014, 15:04:36) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import urllib2, m3u8, urlparse
>>> url='https://devimages.apple.com.edgekey.net/streaming/examples/bipbop_4x3/bipbop_4x3_variant.m3u8'
>>> text=urllib2.urlopen(url).read()
>>> m=m3u8.M3U8(content=text, base_uri=url)
>>> m.playlists[0].absolute_uri
'https://devimages.apple.com.edgekey.net/streaming/examples/bipbop_4x3/bipbop_4x3_variant.m3u8/gear1/prog_index.m3u8'
```

My expectation is that m.playlist[0].absolute_uri would be "https://devimages.apple.com.edgekey.net/streaming/examples/bipbop_4x3/gear1/prog_index.m3u8", as you would get from urlparse.urljoin, which is the actual location of the playlist that a video player would retrieve (a relative path derived from the original uri, NOT a sub-path of the original uri). Continuing on, that url that I was expecting does in fact succeed:

```
>>> urlparse.urljoin(url, m.playlists[0].uri)
'https://devimages.apple.com.edgekey.net/streaming/examples/bipbop_4x3/gear1/prog_index.m3u8'
>>> subtext=urllib2.urlopen(urlparse.urljoin(url, m.playlists[0].uri)).read()
>>> len(subtext)
6701
>>> sub_m=m3u8.M3U8(subtext)
>>> len(sub_m.segments)
181
```

Whereas the absolute_uri from the playlist fails:
```
>>> m.playlists[0].absolute_uri
'https://devimages.apple.com.edgekey.net/streaming/examples/bipbop_4x3/bipbop_4x3_variant.m3u8/gear1/prog_index.m3u8'
>>> m.playlists[0].uri
'gear1/prog_index.m3u8'
>>> subtext=urllib2.urlopen(m.playlists[0].absolute_uri)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #53, opened 2015-07-28
Fixed https://github.com/globocom/m3u8/issues/42 plus other minor problems

GrumpyOldTroll:
Unknown AttributeNames should be ignored, according to https://tools.ietf.org/html/draft-pantos-http-live-streaming-16#section-6.3.1.  Observed BANDWIDTH included in EXT-X-MEDIA tag from ted.com (e.g. http://hls.ted.com/talks/2297.m3u8), which is not defined, but caused an exception. Fixed that with the **extras.

Also changed semantics when an absolute path without a netloc is used in a url, so that base_url will behave like a urljoin, using base_url netloc and the given url's absolute path. Note the change to expected_ts1_abspath in test_load_should_create_object_from_uri_with_relative_segments in test_loader.py. Also impacted parsing of the same video from ted.com.

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -131,6 +131,10 @@ class M3U8(object):
         else:
             self.data = {}
         self._base_uri = base_uri
+        if self._base_uri:
+            if not self._base_uri.endswith('/'):
+                self._base_uri += '/'
+
         self._initialize_attributes()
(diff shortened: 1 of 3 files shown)


leandromoreira:
Thanks a lot @GrumpyOldTroll

PR #54, opened 2015-08-06
Average bandwidth

cristina0botez:
Added the AVERAGE-BANDWIDTH attribute for  EXT-X-STREAM-INF (https://tools.ietf.org/html/draft-pantos-http-live-streaming-16#section-4.3.4.2).


leandromoreira:
Awesome thanks so much @cristina0botez

PR #64, opened 2015-10-15
Mote python 3 support.

audetto:
Now all tests pass in python 3 as well.

Andrea

This is to fix

TypeError: unorderable types: NoneType() > int()

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -222,7 +222,7 @@ class M3U8(object):
         output = ['#EXTM3U']
         if self.is_independent_segments:
             output.append('#EXT-X-INDEPENDENT-SEGMENTS')
-        if self.media_sequence > 0:
+        if self.media_sequence:
             output.append('#EXT-X-MEDIA-SEQUENCE:' + str(self.media_sequence))
         if self.allow_cache:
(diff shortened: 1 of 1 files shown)


audetto:
Not sure what is happening here.
This fixes the code in python 3 so all the tests pass.

leandromoreira:
yeah I see, @audetto it seems that's something related to coverage 
` "/home/travis/virtualenv/python3.2.5/lib/python3.2/site-packages/coverage/annotate.py", line 85
    dest.write(u'  ')`

audetto:
coverage at line 85 of annotate.py

https://github.com/nedbat/coveragepy/blob/master/coverage/annotate.py#L85

uses a feature only available in python 3.3+

u'XXXXX' is a Unicode literal string

https://docs.python.org/3/whatsnew/3.3.html

The doc of coverage says
https://pypi.python.org/pypi/coverage

**Coverage.py runs on CPython 2.6, 2.7, 3.3, 3.4 or 3.5, PyPy 2.6, and PyPy3 2.4.**

You probably need to drop support for 3.2

leandromoreira:
@audetto cool, I'll do it. Thanks =) :beers:

PR #65, opened 2015-10-20
Fix bugs in resolution attribute and dumps function.

somsak:
- Add support for double quote in resolution attributes
- Fix how segment represent itself (__str__)
- Bump version up


leandromoreira:
Thanks a lot @somsak :cake:

leandromoreira:
@somsak :beers:  https://pypi.python.org/pypi/m3u8/0.2.8

PR #70, opened 2016-07-04
Multi media fixes

Prendo93:
Currently master playlists print the media type once for every media in a playlists media array:

PR contains:
f6c39e2
 New Test Playlist: MULTI_MEDIA_PLAYLIST 
 Test illustrating that playlist->loads->dumps != playlist
7a4a8e9
 Small mod to __str__ function of Playlist model so each media type only prints once.


leandromoreira:
Thank you very much @Prendo93

PR #71, opened 2016-07-07
add timeout parameter for load function

hitzzc:
I met a case that my project hang on the load function because there is no timeout parameter. I think add timeout parameter into load function is better


leandromoreira:
Thanks @hitzzc I'd ask you to provide test case for when using timeout though.

hitzzc:
Hi @leandromoreira , sorry for my late reply. 
Test case has been added.

leandromoreira:
thank you, so much, do you need a new version with   this?

hitzzc:
Hi @leandromoreira , I don't need a new version with this. Thanks

PR #74, opened 2016-08-08
Extract SCTE35 metadata on every segments

birme:
Some HLS encoders (such as Elemental) provides the functionality to provide SCTE35 metadata on each segment in a period. This pull request adds the functionality to extract the SCTE35 metadata and duration.


leandromoreira:
Thanks a lot @birme

leandromoreira:
do you need a new build with this @birme ?

birme:
That would be great!

leandromoreira:
done! :) thanks again

PR #75, opened 2016-08-10
Update README; multiple keys are supported now

Alexqw:
At least according to issue 1 and the release notes of 0.2.1.


leandromoreira:
thanks

PR #78, opened 2016-08-24
Fix ParseError exception

ziima:


leandromoreira:
@ziima thank you very much :)

PR #81, opened 2016-09-27
SCTE35 in CUE-OUT

birme:
I have discovered that some HLS flavours (Elemental) provides actually the first SCTE35 marker and duration on the EXT-CUE-OUT tag and not the EXT-CUE-OUT-CONT which by only looking for CUE-OUT-CONT made us be a segment off.


leandromoreira:
https://trac.ffmpeg.org/ticket/3356

birme:
Great! When is the next release scheduled?

leandromoreira:
i'll try today, thank you very much.

PR #83, opened 2016-10-04
Changed implementation from single key to multiple keys in M3U8

bertothunder:
The new list of keys is extracted (aside from keys stored in segments). This allows more flexibility in the handling of all possible encryption scenarios. M3U8 documentation shows the possible scenarios and the change.

Added new tests for the change on keys, and commented out other ones temporarily (to be deleted later?).
Mixins have been moved to an external file for a more clear code and better organization.

The list of keys is not a set of UNIQUE keys, this will be changed in a later release with shared keys (instead of create them with each segment, which brings different instances and blocks the uniqueness I'm looking for). Also, that will allow to get list of segments per key.

Code has been sanitized with autopep8. Updated .gitignore for virtualenvs too.


m3u8/model.py
```diff
@@ -137,35 +147,33 @@ def __init__(self, content=None, base_path=None, base_uri=None, strict=False):
...
                                       for params in self.data.get('segments', []) ])
-
+        self.keys = [ key for key in set(segment.key for segment in self.segments) ]
         for attr, param in self.simple_attributes:
             setattr(self, attr, self.data.get(param))
 
         self.files = []
-        if self.key:
-            self.files.append(self.key.uri)
+        for key in self.keys:
```
leandromoreira: We could try to do list comprehension here and then a filter `L = filter(None, originalList)` and apply unique.

leandromoreira:
Thank you very much @bertothunder

bertothunder:
On 4 October 2016 02:28:49 BST, Leandro Moreira <notifications@github.com> wrote:
>leandromoreira commented on this pull request.
>
>Also we should check why some test have failed.

Which tests have failed? I get 92 tests running with no problem. Have they failed in travis?

I have to have a look at the results, noticed build failed for python 2.6 and 3.x, probably imports, didn't use any specific language trick that could affect it.

>
>>          for attr, param in self.simple_attributes:
>             setattr(self, attr, self.data.get(param))
> 
>         self.files = []
>-        if self.key:
>-            self.files.append(self.key.uri)
>+        for key in self.keys:
>
>We could try to do list comprehension here and then a filter `L =
>filter(None, originalList)` and apply unique.

The coding style in python is changing to use simpler calls (ie replace 'reduce' by a for loop) to increase readibility; not sure if it would increase performance either, given the tipically small number of elements in the list, that's why I used the 'for' loop. Up to you ;-).

Thanks
Alberto

bertothunder:
I've fixed the import errors in Python 2.6, and cleaned a couple of ugly prints used in debugging. 

The build still fails in Python 3.x because Key class is not hashable (does not have __hash__ method), so in python 3.x can't be used from the set for comparison. I'll think about it and fix it later.

leandromoreira:
Thanks once again, I am fine with the `for` 👍  now I'll wait for the python3 fix.

bertothunder:
Leandro, I've fixed the build issues in all versions of Python. 

'keys' property is now updated to be a unique set of keys used all across the segments. Also, I've added a method by_key() to SegmentList, allowing to filter segments encrypted by a given key, and allowing something new: to update the key associated with them!

Please check documentation in model.py and README.rst.

New tests added, and old ones have been restored.

leandromoreira:
Thank you very much @bertothunder

leandromoreira:
Do you need a build with this changes?

bertothunder:
No, it's ok for me, thank you ;)

leandromoreira:
Pretty neat your contribution (test, code and doc), thank you so much!

bertothunder:
Please check the README.rst description for multiple keys, you may like to express it in some other way.

leandromoreira:
I'm thinking to bumb up the version to 0.3.0 because we changed the interface from `key` to `keys` what do you think?

leandromoreira:
@bertothunder I checked (it's fantastic 🍻  ) and I saw a example that, at first, didn't make sense to me, why is this `segmk1 = m3u8_obj.segments.by_key(None)` necessary? (I saw it on the README.rst)

bertothunder:
For the first question, give me 2 secs, I found a ugly formatting of the code done by autopep8 in parser.py, where code is broken in this way: 

   function.call(
      parameters)

 instead of :   function.call(parameters)

 I'll fix it and create another pull request.

For the second questions: let's imagine you have a complex m3u8 where first segments are unencrypted but, at some point in the time window, encryption starts, so you may have something like this (quite extreme example, time window should never be so big from an encoder):

#EXTM3U
#EXT-X-TARGETDURATION:10
#EXT-X-VERSION:3
#EXT-X-MEDIA-SEQUENCE:0
#EXTINF:10.0,
file0.ts
#EXTINF:10.0,
file1.ts
#EXTINF:10.0,
file2.ts
#EXT-X-KEY:METHOD=AES-128,URI="https://priv.example.com/key.php?r=52",  \
IV=0x9c7db8778570d05c3177c349fd9236aa
#EXTINF:10.0,
file3.ts
#EXTINF:8.0,
file4.ts
#EXT-X-KEY:METHOD=AES-128,URI="https://priv.example.com/key.php?r=53",  \
IV=0xc055ee9f6c1eb7aa50bfab02b0814972
#EXTINF:10.0,
file5.ts
#EXTINF:10.0,
file6.ts
#EXTINF:10.0,
file6.ts


In this example, m3u8.keys() will return 3 keys in the list: None (for the first segments), key1 and key2 (let's call this way for the sake of simplicity).

If you want to get the segments encrypted by last key, you would do:
m3u8.segments.by_key( m3u8.keys[-1] )  

as shown in the readme.rst. The same would apply for the first segments (with key None):

m3u8.segments.by_key( m3u8.keys[0] )

But in this case, this is comparable to 

m3u8.segments.by_key(None)

because there is a 'None' key in the list of keys. Is that clear?

Actually, thinking about it, we could create a KeyList class, with filter methods:
- by_uri
- by_iv
- by_method

This way, it would be possible to make:

  key1 = m3u8.keys.by_uri("https://priv.example.com/key.php?r=53")
  segm = m3u8.segments.by_key(key1)

Maybe something you are interested on implementing yourself? Not difficult at all.

bertothunder:
I've created the new pull request with the ugly broken lines fixed. 

Regarding the new version, it makes all sense with the new API.

leandromoreira:
>  Is that clear?

Yes.

bertothunder:
@leandromoreira  Sorry, the question was not intended to be harsh, just not being english my native language brings sometimes the wrong way of expressing questions. I've updated the comment above, apologies.

leandromoreira:
easy ;)

daw134:
I've got an M3U8 with multiple Keys, every 10 segments theres a new URI, same URL but the KID changes at the end and the IV changes every 10 segments too obviously. When i print key.uri it prints ALL of the URLs but 10 of them are the same and I only need one of each, how can i get it to print / return only one of each URL?

Issue #85, opened 2016-10-05
Support for Envivio flavored HLS markers


birme:
Envivio encoders provide a bit different flavour of HLS SCTE 35 cue markers. Would be great to have support for these.

PS: I will provide a PR for this

leandromoreira:
Can we close this one?

birme:
Sure!

PR #86, opened 2016-10-05
Support for Envivio flavoured SCTE35 cue markers

birme:
Provide support for #85


m3u8/parser.py
```diff
@@ -280,15 +284,20 @@ def _parse_cueout(line, state):
         state['current_cue_out_duration'] = res.group(1)
         state['current_cue_out_scte35'] = res.group(2)
 
-
 def _parse_cueout_start(line, state, prevline):
     param, value = line.split(':', 1)
     state['current_cue_out_duration'] = value
+    # Try Elemental flavor first
```
leandromoreira: How about to use two functions to parse this value:

```python
def _cueout_elemental(prevline):
    return ...


def _cueout_envivio(prevline):
    return ...


# and then

state['x'] = _cueout_elemental(prev) or _cueout_envivio(prev)
```

What do you think? My intention was to make it clear.

tests/test_model.py
```diff
@@ -80,6 +80,13 @@ def test_segment_scte35_attribute():
     assert segments[9].cue_out == False
     assert segments[4].scte35 == '/DAlAAAAAAAAAP/wFAUAAAABf+//wpiQkv4ARKogAAEBAQAAQ6sodg=='
 
+def test_segment_envivio_scte35_attribute():
+    obj = m3u8.M3U8(playlists.CUE_OUT_ENVIVIO_PLAYLIST)
+    segments = obj.segments
+    assert segments[3].cue_out == True
+    assert segments[4].scte35 == '/DAlAAAENOOQAP/wFAUBAABrf+//N25XDf4B9p/gAAEBAQAAxKni9A=='
+    assert segments[5].scte35 == '/DAlAAAENOOQAP/wFAUBAABrf+//N25XDf4B9p/gAAEBAQAAxKni9A=='
+    assert segments[7].cue_out == False
```
leandromoreira: I also would provide an elemental test, since this PR provides both implementations.

m3u8/parser.py
birme: Agree, that would be more clear.

tests/test_model.py
birme: Sure!

m3u8/parser.py
```diff
@@ -280,15 +284,29 @@ def _parse_cueout(line, state):
...
+    param, value = line.split(':', 1)
+    res = re.match('.*DURATION=(.*),.*,CUE="(.*)"', value)
+    if res:
+        state['current_cue_out_duration'] = res.group(1)
+        state['current_cue_out_scte35'] = res.group(2)
+        return True
+    else:
+        return False
 
+def _parse_cueout_start(line, state, prevline):
```
leandromoreira: Cool, I liked it! I'd suggest (but this is only kinda of a stylish thing, more functional instead of dealing with state inside that function you return the values), be cool to skip this one:

```python
def _cueout_elemental(...):
      param, value = line.split(':', 1)
      res = re.match('.*EXT-OATCLS-SCTE35:(.*)$', prevline)
      if res:
         return (res.group(1), value)
     else:
         return None

# ... the same for the other parser and

_cueout_state = _cueout_elemental(line, state, prevline) or _cueout_envivio(line, state, prevline)
if _cueout_state:
  state['current_cue_out_duration'] = _cueout_state(1)
  state['current_cue_out_scte35'] = _cueout_state(2)
```

tests/test_model.py
leandromoreira: Nice 👍

m3u8/parser.py
birme: That is of course a nice idea. I like it. Will fix

birme:
If you accept this and merge this it would be great to get a package built for this. Thanks in advance!

leandromoreira:
Thank you very much @birme I'll release a new version today.

leandromoreira:
done https://github.com/globocom/m3u8/releases/tag/0.3.0

Issue #58, opened 2015-10-09
if EXT-X-KEY:METHOD=NONE, get TypeError: __init__() takes at least 4 arguments (3 given)


sloh:
in model.py > Class Segment() > __init__() for the following line:
```python
self.key = Key(base_uri=base_uri,**key) if key else None
```

if a segment has "#EXT-X-KEY:METHOD=NONE", it is missing the uri parameter needed for class Key(), its init function:
```python
def __init__(self, method, uri, base_uri, iv=None, keyformat=None, keyformatversions=None):
```

I believe uri should also be an optional parameter according to this documentation:
```
URI

   The value is a quoted-string containing a URI that specifies how to
   obtain the key.  This attribute is REQUIRED unless the METHOD is
   NONE.
```
http://tools.ietf.org/html/draft-pantos-http-live-streaming-16#section-4.3.2.4

bertothunder:
I can confirm URI should be optional, if METHOD=NONE then most of encoders or packagers won't add URI onto it.

davidjameshowell:
I can confirm the same. Encountered this most recently with content that I had been using that has  METHOD=None. I had to get the script working, so I just went through and edited it to return none for the key for the time being until a fix is rolled.

leandromoreira:
@davidjameshowell try now with @birme fixes (you need to checkout the master)

bertothunder:
@leandromoreira fix from @birme is wrong: the URI is **optional**, but his fix forces URI to be empty string (not None), so the fix drops the URI even if provided.

Check my commit, it fixes the issue and brings a new test for this.

bertothunder:
So my fix is wrong as well as the original fix for this issue. As the uri must NOT be present (as pointed out by Bradley). 

So, as pointed out by the author, uri must be none in the constructor to avoid the original error.

leandromoreira:
Now, I'm confused, should I revert some commits?

birme:
I understand what you mean and will fix. If method is NONE uri is optional but if provided it should remain intact

birme:
Disregard this comment. Saw the other thread. According to spec if method is none no following attributes must be present. So go ahead and revert my fix as the original problem it solves was not a problem...

bertothunder:
Well, thinking about it, I did actually close my pull request too soon.

[...]

PR #87, opened 2016-10-10
Suggested fix for #58: If key-method is None the key uri should not be mandatory

birme:
Suggested fix for issue #58 to handle the case when key-method is NONE the key URI should not be mandatory.


leandromoreira:
cool, thank you very much! :) @birme you're doing an amazing work.

birme:
Glad I can help!

PR #88, opened 2016-10-10
Fixes commit for issue-58, where key was forced to be empty string even if provided.

bertothunder:
Commit for issue-58 was wrong; when the key method is NONE, according to RFC, URI becomes optional. But the patch for this issue makes URI to be empty string no matter what.

This issue enforces the key URI is None **only** is it's not provided.

Existing test has been ammended, and a new test has been added with a key being provided with method NONE and URI "", that will be kept as-is.


bradleyfalzon:
According to RFC: https://tools.ietf.org/html/draft-pantos-http-live-streaming-20#section-4.3.2.4

```
      An encryption method of NONE means that Media Segments are not
      encrypted.  If the encryption method is NONE, other attributes
      MUST NOT be present.
```

You maybe referencing:

```
      The value is a quoted-string containing a URI that specifies how
      to obtain the key.  This attribute is REQUIRED unless the METHOD
      is NONE.
```

If METHOD is NONE, URI isn't optional - it MUST NOT be present.

bertothunder:
Yep, you are right, I got confused, checked it again. Then the original issue and the patch for it must be reverted as well.

PR #90, opened 2016-10-11
Partial revert (maintaining the added unit tests) in the fix for #58

birme:
Based on the discussion in #58 and in light of that the RFC stating that URI is only required if key method is not NONE I have reverted the fix. And actually other attributes MUST NOT be present if method is NONE. That part is not enforced in this commit. It may be value to keep the added test cases already merged, and why I created a new PR with the revert. It is up to you if you want to use this or just revert the old PR.

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -412,7 +412,7 @@ class Key(BasePathMixin):
 
     '''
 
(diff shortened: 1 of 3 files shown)


leandromoreira:
Thank you very much sir :)

Issue #72, opened 2016-07-14
BANDWIDTH format change on Twitch playlists


ilyalissoboi:
It appears that Twitch.tv is breaking/changing specification by producing master playlists that have bandwidth described by a floating point number instead of an integer.

An example of such a playlist can be seen [here](http://pastebin.com/TUUT1YvZ).

Is it possible to accommodate that change by changing the expected bandwidth format in the module?

leandromoreira:
> BANDWIDTH
>
 >     The value is a decimal-integer of bits per second.  It represents the peak segment bit rate of the Variant Stream.

I think we should let it parsed to `int` as is stated by the ["official spec"](https://tools.ietf.org/html/draft-pantos-http-live-streaming-19)

leandromoreira:
@globocom/videos3 any thoughts?

flavioribeiro:
based on the example provided by @ilyalissoboi we can receive as float and cast to int

PR #91, opened 2016-10-11
Proposed fix for #72

birme:
This pull request suggests a fix to handle non-conformant variant playlists where BANDWIDTH is denoted as a float and not a decimal-integer. Accepting this PR would make the parser a bit less strict so it is up to the maintainer(s) of this package to decide whether that path should be taken or not.


leandromoreira:
What do you think @flavioribeiro ? Here's some context, BANDWIDTH [by definition is an INT](https://tools.ietf.org/html/draft-pantos-http-live-streaming-20#section-4.3.4.2) but some "vendors" specifically twitch is using it as floating point, for me it's okay.

I mean, it doesn't break old clients and still offer this workaround.

Issue #94, opened 2017-01-13
Add the ability to set headers for URI


baco:
Some sites require setting some headers like ``Referer`` or ``Cookie``, even the ``UserAgent`` sometimes, to access the the content and don't be filtered by the webserver. Using ``m3u8.load()`` won't allow you to setup anything of the sort.

Of course, it is possible to bypass the issue by making the request by hand using ``urllib.request`` module and then ``m3u8.loads()`` but it adds extra  complexity and the resulting m3u8 objects (playlists or segments) lack of some attributes like ``.absolute_uri`` then.

One possible solution is to add an extra argument to ``m3u8.load()``, for example ``headers={}``, and when set to anything different than that, make the request using ``Request()`` and then ``urlopen()`` inside the ``m3u8.load()`` method.

leandromoreira:
Hi @baco thank you for commenting, so what do you really want to have is the ability to add `header` for all the requests?

baco:
Yes, sorry; I've asked for `data` field but `header` was the one I am needing right now. I am updating the issue in accordance.

leandromoreira:
@baco I'll close this one since it's the same as https://github.com/globocom/m3u8/issues/82 do you agree?

baco:
Agreed ☺️

Issue #82, opened 2016-10-02
Custom HTTP Headers


PhilippeR:
Hi,
Is it possible to customize request HTTP Headers ?  
Tks.
Philippe

leandromoreira:
I'm sorry, can you give an example of what you want to do?

PhilippeR:
Add custom http headers or override standard HTTP header like user Agent in http request.
We can imagine m3u8_obj=m3u8.load(value, headers) 

Exemples: 
X-Playback-Session-Id: 8ECAD943-8CC6-40B5-8550-E64EE579BB0D
X-MYCUSTOM-HEADER: Let's Go
Connection: keep-alive
Accept: */*
User-Agent: AppleCoreMedia/1.0.0.14A456 (iPad; U; CPU OS 10_0_2 like Mac OS X; fr_fr)
Accept-Language: fr-fr
Accept-Encoding: identity
Connection: keep-alive

leandromoreira:
It isn't possible to this now but it sounds like a fair feature and it needs to be [compatible for p2 and p3](https://github.com/globocom/m3u8/tree/master/m3u8#L13-L19).

```
# this solves the issue while we're using urllib2
import urllib2
request = urllib2.Request("http://www.google.com", headers={"Accept" : "text/html"})
contents = urllib2.urlopen(request).read()
```

PhilippeR:
Thanks ;-)
Believe me, It makes sense to choose them sometimes.,

mauricioabreu:
Support for `custom headers` was added in #95 
Closing.

PR #95, opened 2017-01-19
Add support for headers to URL opener

baco:
Added the ability to build the request for an URL specifying more
arguments such as headers.  With this option cookies, referrer and
user-agent may be passed to access the content of URLs.

Closes: #94, #82


leandromoreira:
Thank you @baco I'd ask you though to add some tests and also to fix the current [build](https://travis-ci.org/globocom/m3u8/jobs/193399430)

PR #96, opened 2017-01-23
using a valid m3u8 url

redstoneleo:


leandromoreira:
Thank you very much but I'd like to leave as it is, mostly because we also could be using http://www.streambox.fr/playlists/x36xhzz/x36xhzz.m3u8 or even one from mp4example.

PR #100, opened 2017-03-10
Bug fix

josephwahba89:
Some of the manifests contain key method = None. This causes the m3u8 object creation fails

### DIFF
--- a/m3u8/parser.py
+++ b/m3u8/parser.py
@@ -106,9 +106,10 @@ def parse(content, strict=False):
 
         elif line.startswith(protocol.ext_x_key):
             key = _parse_key(line)
-            state['current_key'] = key
-            if key not in data['keys']:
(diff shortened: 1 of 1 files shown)


leandromoreira:
Thank you very much  @josephwahba89 ! Is it None or NONE? also, can you provide a test case ?

leandromoreira:
Looking at the [failed build](https://travis-ci.org/globocom/m3u8/jobs/209792104), it seems that we already take care of this...

ebs111:
Hi
I need some help please
When i use import m3u8 in my code i get that the moudle is missing and i cannot find the m3u8.py file .
How to solve it?
Thanks

leandromoreira:
@ebs111 try to see if you didn't create a file named m3u8.py and tried to import m3u8 module, aka there might be a naming conflict.

Issue #102, opened 2017-04-08
Doesn't parse #EXT-X-MAP attribute in m3u8


denex:
Some playlists contain #EXT-X-MAP:URI="fileSequence0.mp4" and this attribute doesn't processed.

>    The EXT-X-MAP tag specifies how to obtain the Media Initialization
>    Section (Section 3) required to parse the applicable Media Segments.
>    It applies to every Media Segment that appears after it in the
>    Playlist until the next EXT-X-MAP tag or until the end of the
>    playlist.
> 
>    Its format is:
> 
>    #EXT-X-MAP:<attribute-list>

https://tools.ietf.org/html/draft-pantos-http-live-streaming-21#section-4.3.2.5

leandromoreira:
https://github.com/globocom/m3u8/pull/103#pullrequestreview-32380019

PR #103, opened 2017-04-10
Added EXT-X-MAP:URI attribute

denex:
Fixed issue https://github.com/globocom/m3u8/issues/102
Added EXT-X-MAP:URI attribute to parser


leandromoreira:
Thank you so much :)

leandromoreira:
Maybe we should document at README all supported tags we've implemented.

PR #106, opened 2017-09-11
Add support CLOSED-CAPTIONS for StreamInfo

Intar:
About problem on iOS: https://github.com/jwplayer/jwplayer/issues/1060
About this tag on draft: https://tools.ietf.org/html/draft-pantos-http-live-streaming-23
On apple dev site: https://developer.apple.com/library/content/qa/qa1801/_index.html


Intar:
I try to find some real case for test, in our case we have only NONE in playlist.

leandromoreira:
Thank you very much @Intar

Issue #107, opened 2017-09-12
m3u8.dump() throws when relative filepath does not contain any directories


budziq:
```python
In [39]: master.dump("master.m3u8")
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-39-a612dd8dcc84> in <module>()
----> 1 master.dump("master.m3u8")

/usr/local/lib/python3.5/dist-packages/m3u8/model.py in dump(self, filename)
    271         Saves the current m3u8 to ``filename``
    272         '''
--> 273         self._create_sub_directories(filename)
    274
    275         with open(filename, 'w') as fileobj:

/usr/local/lib/python3.5/dist-packages/m3u8/model.py in _create_sub_directories(self, filename)
    279         basename = os.path.dirname(filename)
    280         try:
--> 281             os.makedirs(basename)
    282         except OSError as error:
    283             if error.errno != errno.EEXIST:

/usr/lib/python3.5/os.py in makedirs(name, mode, exist_ok)
    239             return
    240     try:
--> 241         mkdir(name, mode)
    242     except OSError:
    243         # Cannot rely on checking for EEXIST, since the operating system

```

It fails due to `_create_sub_directories` assumption that a filepath will have a `basename` which is not a case for relative paths without directories such as "master.m3u8". This might be an edge case but it has cost me few minutes of debugging ;)

PR #108, opened 2017-09-12
fix for m3u8.dump() failing on relative paths without directories

budziq:
fixes https://github.com/globocom/m3u8/issues/107


leandromoreira:
Thank you very much @budziq

Issue #110, opened 2017-10-02
#EXT-X-MEDIA uri is optional


eventh:
I have a m3u8 master playlist with an ext-x-media tag without any uri attribute.
It is an audio group where the uri is found in an #EXT-X-STREAM-INF.

The parser fails when I use base_path, because setting base_path calls
os.path.dirname(self.uri), when self.uri is None.

File "/usr/local/lib/python3.6/site-packages/m3u8/mixins.py", line 33, in base_path
    return os.path.dirname(self.uri)

According to the specification, the uri in media tag is optional when the type is audio or video:
https://tools.ietf.org/html/draft-pantos-http-live-streaming-21#section-4.3.4.2.1

#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="audio-aacl-312",NAME="English",LANGUAGE="en",AUTOSELECT=YES,DEFAULT=YES,CHANNELS="2"
 
#EXT-X-STREAM-INF:BANDWIDTH=364000,AVERAGE-BANDWIDTH=331000,CODECS="mp4a.40.2",AUDIO="audio-aacl-312",SUBTITLES="textstream"
ch001-audio_312640_eng=312000.m3u8

[index.txt](https://github.com/globocom/m3u8/files/1349168/index.txt)

eventh:
I would be happy to contribute a fix, but I'm not sure what would be the preferred way of solving this issue. Should the uri be made optional, or should it be read from the ext-x-stream-inf?

PR #111, opened 2017-10-03
Setting base_path when uri is None is allowed now, and does nothing.

eventh:
Since uri is optional, and setting base_path on M3U8 objects
sets base_path on all media/playlist/key objects of the M3U8 object,
the base_path must check first if uri is actually set.

Added two tests that fail without this change, but passes after.

Solves #110 

Signed-off-by: Even Thomassen <even.thomassen@noriginmedia.com>


leandromoreira:
Thank you very much @eventh !!!

PR #112, opened 2017-10-15
Don't fail if title has a comma in it

guthypeter:
m3u8 failed, when there was a comma in the title, ex
 #EXTINF:295,Test 123, part 14

We can only split into max 2 parts.

### DIFF
--- a/m3u8/parser.py
+++ b/m3u8/parser.py
@@ -174,7 +174,7 @@ def _parse_key(line):
 
 
 def _parse_extinf(line, data, state, lineno, strict):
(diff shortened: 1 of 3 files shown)


leandromoreira:
Thank you @guthypeter can you add a test to cover this situation? (where titles contains commas)

leandromoreira:
https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.2.1

leandromoreira:
Thank you very much 👏 👏 👏 👏 👏 👏 👏 👏 👏

PR #113, opened 2017-11-08
Add time offset support

blokfyuh:
Added support for #EXT-X-START parameter.


leandromoreira:
https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.5.2

blokfyuh:
@leandromoreira Could you elaborate ? Am I missing something ?

leandromoreira:
Thanks so much @blokfyuh while I reading the [official docs](https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.5.2) I noticed some things:

The tag `#EXT-X-START:<attribute-list>` returns an attribute list value being one of them a `TIME-OFFSET` which is a signed-decimal-floating-point number of seconds therefore we should use `_parse_simple_parameter(line, data, float)` instead of the simple `_parse_simple_parameter(line, data)`. The other one is `PRECISE` which is the value is an enumerated-string; valid strings are YES and NO.

blokfyuh:
Absolutely you're right. I'll do a quick fix, let me know if it's ok for you.

blokfyuh:
@leandromoreira I used a different approach, let me know what you think about it.
Question: Do you prefer to handle incorrect values at the parser level or at the model one (i.e `class Start(object)` ?

leandromoreira:
Cool @blokfyuh I tend to prefer to validate at model level =) 

Can you also add some test cases for this tag?

1. `#EXT-X-START:TIME-OFFSET=-2` (PRECISE defaults to NO, negative value (valid))
2. `#EXT-X-START:TIME-OFFSET=10.5,PRECISE=YES` (with precise, double value)
3. `#EXT-X-START:TIME-OFFSET=145` (without precise, int value)

blokfyuh:
@leandromoreira I added the unit tests as you've asked.
Regarding your "PRECISE defaults to NO", I didn't add it to the code because the official documentation says that it's the client hls responsability to consider NO by default if it's missing. But we shouldn't turn a user M3U8 file without PRECISE to an output with PRECISE=NO. Don't you think ?

tests/test_model.py
```diff
@@ -712,6 +715,21 @@ def test_segment_map_uri_attribute_with_byterange():
     obj = m3u8.M3U8(playlists.MAP_URI_PLAYLIST_WITH_BYTERANGE)
     assert obj.segment_map['uri'] == "main.mp4"
 
+
+def test_start_with_negative_offset():
+    obj = m3u8.M3U8(playlists.SIMPLE_PLAYLIST_WITH_START_NEGATIVE_OFFSET)
+    assert obj.start.time_offset == -2.0
+    assert obj.start.precise is None
```
leandromoreira: Do you think it's better to use 'NO' or `None`?

leandromoreira: I read your idea, yeahhh ;) I agree with you let's leave it `None`.

leandromoreira:
> the client hls responsability to consider NO by default if it's missing. But we shouldn't turn a user M3U8 file without PRECISE to an output with PRECISE=NO. Don't you think ?

Yes, you're totally right, let's use `None` to avoid an unnecessary dump content.

PR #114, opened 2017-11-18
Now dumps EXT-X-MAP:URI= to output

jaredLunde:


leandromoreira:
Thank you very much @jaredlunde can you please add a test to this case?

jaredlunde:
Sure thing - will push that tomorrow.

mauricioabreu:
@jaredLunde sorry for the long time to reply.
I added a new PR including the `URI` and `byterange` (with the missing tests)
Thank you for your efforts to make this library better!

PR #115, opened 2017-11-24
No longer required a segment to be mark as a discontinuity in order include program date time.

cameronmaske:
Hey there, 
I'm not sure if this is a bug, or part of the HLS spec requirement (I tried googling around, but didn't come across anything relevant). 

I came across this manipulating some m3u8 which have program dates for each segments, e.g.
Here is an example of the loading and dumping, that cause the #EXT-X-PROGRAM-DATE-TIME to be wiped on segments. 

Input (original) 
```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-TARGETDURATION:7
#EXT-X-MEDIA-SEQUENCE:0
#EXT-X-PLAYLIST-TYPE:EVENT
#EXTINF:6.473522,
#EXT-X-PROGRAM-DATE-TIME:2017-11-24T12:43:39.000+0000
003.ts
#EXTINF:5.866667,
#EXT-X-PROGRAM-DATE-TIME:2017-11-24T12:43:45.474+0000
004.ts
#EXT-X-ENDLIST
```
Output (output dumped)
```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-TARGETDURATION:7
#EXT-X-PROGRAM-DATE-TIME:2017-11-24T12:43:39+00:00
#EXT-X-PLAYLIST-TYPE:EVENT
#EXTINF:6.473522,
003.ts
#EXTINF:5.866667,
004.ts
#EXT-X-ENDLIST
```

Thoughts?

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -360,9 +360,9 @@ class Segment(BasePathMixin):
 
         if self.discontinuity:
             output.append('#EXT-X-DISCONTINUITY\n')
-            if self.program_date_time:
-                output.append('#EXT-X-PROGRAM-DATE-TIME:%s\n' %
-                              format_date_time(self.program_date_time))
+        if self.program_date_time:
+            output.append('#EXT-X-PROGRAM-DATE-TIME:%s\n' %
+                          format_date_time(self.program_date_time))
         if self.cue_out:
             output.append('#EXT-X-CUE-OUT-CONT\n')
         output.append('#EXTINF:%s,' % int_or_float_to_string(self.duration))


leandromoreira:
Looking at the [specs](https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.2.6):
> The EXT-X-PROGRAM-DATE-TIME tag associates the first sample of a Media Segment with an absolute date and/or time.
I think we can safely assume that is part of the hls spec.

cameronmaske:
@leandromoreira Sorry, my comment was confusing. I meant, specwise, if `EXT-X-PROGRAM-DATE-TIME` can only be used in segments, if used in conjunction with `EXT-X-DISCONTINUITY` and can, or if can stand alone.

leandromoreira:
When I read the spec my understanding was: "We should apply the tag `EXT-X-PROGRAM-DATE-TIME` only for the first media segment or when a `EXT-X-DISCONTINUITY` happens" if that's the case then your input example is wrong. ;/ I also take a look at some HLS generate by elemental, woowza and ffmpeg and notice that they all respect this rule.

IE: https://video-dev.github.io/streams/x36xhzz/x36xhzz.m3u8

igorsobreira:
I'm having the same issue, loading a m3u8 that contains `#EXT-X-PROGRAM-DATE-TIME` but **not** `EXT-X-DISCONTINUITY`. Then dumping will strip out the `#EXT-X-PROGRAM-DATE-TIME`.

The spec says:

>    The server MAY associate an absolute date and time with a Media
   Segment by applying an EXT-X-PROGRAM-DATE-TIME tag to it.  This
   defines an informative mapping of the (wall-clock) date and time
   specified by the tag to the first media timestamp in the segment,
   which may be used as a basis for seeking, for display, or for other
   purposes.  If a server provides this mapping, it SHOULD apply an
   EXT-X-PROGRAM-DATE-TIME tag to every segment that has an
   EXT-X-DISCONTINUITY tag applied to it.
https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-6.2.1

From my understanding, you can provide `EXT-X-PROGRAM-DATE-TIME` for segments to help seeking, display etc. But if you do then you *should* also include `EXT-X-PROGRAM-DATE-TIME` for every segment that has `EXT-X-DISCONTINUITY`

My manifest is coming from AWS MediaPackage

```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-TARGETDURATION:6
#EXT-X-PLAYLIST-TYPE:EVENT
#EXT-X-MEDIA-SEQUENCE:50
#EXT-X-PROGRAM-DATE-TIME:2019-06-10T00:05:00.000Z
#EXTINF:6.000,
manifest_1_50.ts?m=1559946393
#EXT-X-PROGRAM-DATE-TIME:2019-06-10T00:05:06.000Z
#EXTINF:6.000,
manifest_1_51.ts?m=1559946393
#EXT-X-PROGRAM-DATE-TIME:2019-06-10T00:05:12.000Z
#EXTINF:6.000,
manifest_1_52.ts?m=1559946393
#EXT-X-PROGRAM-DATE-TIME:2019-06-10T00:05:18.000Z
#EXTINF:6.000,
manifest_1_53.ts?m=1559946393
#EXT-X-PROGRAM-DATE-TIME:2019-06-10T00:05:24.000Z
#EXTINF:6.000,
manifest_1_54.ts?m=1559946393
#EXT-X-PROGRAM-DATE-TIME:2019-06-10T00:05:30.000Z
#EXTINF:6.000,
manifest_1_55.ts?m=1559946393
#EXT-X-PROGRAM-DATE-TIME:2019-06-10T00:05:36.000Z
#EXTINF:6.000,
manifest_1_56.ts?m=1559946393
#EXT-X-PROGRAM-DATE-TIME:2019-06-10T00:05:42.000Z
#EXTINF:6.000,
manifest_1_57.ts?m=1559946393
#EXT-X-PROGRAM-DATE-TIME:2019-06-10T00:05:48.000Z
#EXTINF:6.000,
manifest_1_58.ts?m=1559946393
#EXT-X-PROGRAM-DATE-TIME:2019-06-10T00:05:54.000Z
#EXTINF:6.000,
manifest_1_59.ts?m=1559946393
#EXT-X-PROGRAM-DATE-TIME:2019-06-10T00:06:00.000Z
#EXTINF:6.000,
manifest_1_60.ts?m=1559946393
#EXT-X-ENDLIST
```

I'm 👍  on this PR

igorsobreira:
Since the PR doesn't have tests I forked and will submit a new one with tests. Turns out this change breaks existing behavior because our `Segment` **always** has `program_date_time` field set, even when `#EXT-X-PROGRAM-DATE-TIME` only exists in the manifest for the first segment

The change as suggested in this PR will break our current `dumps()` because it will add `#EXT-X-PROGRAM-DATE-TIME` for all segments, even those that did not have when we `load`ed

I think the ideal solution would be to only set `Segment.program_date_time` if the segment actually has `#EXT-X-PROGRAM-DATE-TIME`. And maybe provide another field that always have a calculated datetime value, based on previous segments (which is the current behavior of `program_date_time`). But I don't want to break backwards compatibility. 

@leandromoreira thoughts on this?

leandromoreira:
> I think the ideal solution would be to only set Segment.program_date_time if the segment actually has #EXT-X-PROGRAM-DATE-TIME.

I'm ok with this thus the loaded model will be the same when dumped!

igorsobreira:
> I'm ok with this thus the loaded model will be the same when dumped! We should avoid to superficially add stuffs at the dump method or cutting out as the current state of the lib. Btw, nice to see you back 😄

Thanks, great to be back working with video :-). Pretty cool that this project is still active and well maintained! 

I created the pull request #142, added a new field to `Segment` called `current_program_date_time`, (not sold on the same). Let me know what you guys think

igorsobreira:
I believe this one is solved now that #142 was merged

PR #118, opened 2018-03-07
remove redundant urlopen

lqs:


leandromoreira:
Hi @lqs thanks for the contribution but it seems that this change made the [python2.6 stop to work](https://travis-ci.org/globocom/m3u8/builds/350151098?utm_source=github_status&utm_medium=notification)

lqs:
@leandromoreira 
The test failure is caused by the current version of `arrow` library, which doesn't install the required `backports.functools_lru_cache` library on python 2.6 and caused an import error.

leandromoreira:
nice, can you try to update this library in this branch to see if all the test will pass?

lqs:
Added a `pip install backports.functools-lru-cache` in Python 2.x to workaround this issue.

leandromoreira:
But now I'm wondering, is it worth? I mean, we swap a redundant call to some new conditional dependency =/ 

I do value your effort @lqs and again I want to thank you for the contribution but I'd like to let this redundant call instead.

I'm gonna put some others mainteiners on the loop as well @flavioribeiro @jbochi @igorsobreira @hltbra  what do you think?

lqs:
@leandromoreira 
The new dependency is added to workaround the issue caused by the `arrow` library, not about 
 my fix of redundant call.

newmaniese:
With the fixes in #120 the new requirement and the changes in `runtests` aren't needed. I vote we pull those and then this merge will be a good update.

leandromoreira:
@newmaniese are you voting to close this PR or accept it now?

newmaniese:
Accept it without the changes to `runtests`

PR #119, opened 2018-06-20
Parsing logic for #EXT-X-DISCONTINUITY-SEQUENCE

newmaniese:
Updating the parsing and model logic to include DISCONTINUITY-SEQUENCE:
https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.3.3

Includes tests and updated the brief documentation.


newmaniese:
I am looking fixing that Python 2.6 error now and will submit another PR

PR #120, opened 2018-06-21
Arrow 0.10.0 breaks 2.6 support. Pinning tests to arrow==0.9.0 fixes the issue

newmaniese:
This fixes the broken build in #119 and also corrects the issue mentioned in #118 

We might want to consider dropping support for 2.6 as the libraries we use for the tests do the same. I believe there are upgrade routes for all major platforms.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 -r requirements.txt
-arrow
+arrow==0.9.0
 pytest
 bottle
 pytest-cov


leandromoreira:
Thanks @newmaniese if we do drop the support to `2.6` is this still required?

newmaniese:
The tests pass with the latest version of arrow on python 2.7 and 3.6, so I would say dropping support for 2.6 would make this PR obsolete.

Issue #121, opened 2018-07-09
AUDIO, VIDEO, SUBTITLES attributes support in StreamInfo object


mbakholdina:
Is it possible to add AUDIO, VIDEO, SUBTITLES attributes support along with CLOSED-CAPTIONS in StreamInfo object when parsing #EXT-X-STREAM-INF tag? It would be more convenient to get an access to this info directly instead of going through all media objects.

mbakholdina:
Current result (playlist_1.m3u8)
```
Variant 0
http://10.129.128.128:8080/output/test_class_hls/preset_1/dump/video_5/video_5.m3u8
445460
(320, 180)
avc1.420015,mp4a.40.2
"cc"
AUDIO
aud
AUDIO
aud
SUBTITLES
sub
SUBTITLES
sub
```

Expected result
```
Variant 0
[m3u8_parser_problems.zip](https://github.com/globocom/m3u8/files/2468754/m3u8_parser_problems.zip)

http://10.129.128.128:8080/output/test_class_hls/preset_1/dump/video_5/video_5.m3u8
445460
(320, 180)
Avc1.420015,mp4a.40.2
CLOSED-CAPTIONS
cc
AUDIO
aud
SUBTITLES
sub
```

**Problem 1:** "cc" instead of cc without quotation marks
**Problem 2:** CLOSED-CAPTIONS attribute and AUDIO, VIDEO, SUBTITLES attributes are parsed at a different levels. CLOSED-CAPTIONS attribute is available via `m3u8_obj.playlists[i].stream_info.closed_captions `
while other attributes via 
`m3u8_obj.playlists[i].media`
**Problem 3:** m3u8_obj.playlists[i].media returns the list of (type, group_id) declared in EXT-X-MEDIA tags if rendition group is declared in EXT-X-STREAM-INF tag in the corresponding attribute. As a result, we have twice (AUDIO, aud) and twice (SUBTITLES, sub) instead of listing only once.

audetto:
Not sure this is the same question, but I can see to extract all audio information from this playlist

Is this supported?
I only get the Italian audio track at the bottom, but not the English one.

#EXTM3U
#EXT-X-VERSION:4

[...]

PR #126, opened 2018-10-12
Add subtitles, video, and audio group-id refs to StreamInfo

cdunklau:
This partially addresses #121

I'm not sure about this... it seems like this could be a potentially breaking change, and I'm suspicious about the tests all passing with it. It seems to me that this (and #125) should have broken playlist serialization.

Please review.

### DIFF
--- a/tests/playlists.py
+++ b/tests/playlists.py
@@ -96,6 +96,14 @@ http://example.com/with-cc-hi.m3u8
 http://example.com/with-cc-low.m3u8
 '''
 
+VARIANT_PLAYLIST_WITH_VIDEO_CC_SUBS_AND_AUDIO = '''
+#EXTM3U
+#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=7680000,CLOSED-CAPTIONS="cc",SUBTITLES="sub",AUDIO="aud",VIDEO="vid"
+http://example.com/with-everything-hi.m3u8
+#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=65000,CLOSED-CAPTIONS="cc",SUBTITLES="sub",AUDIO="aud",VIDEO="vid"
+http://example.com/with-everything-low.m3u8
+'''
+
 VARIANT_PLAYLIST_WITH_AVERAGE_BANDWIDTH = '''
 #EXTM3U
 #EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=1280000,AVERAGE-BANDWIDTH=1252345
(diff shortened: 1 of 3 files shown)


leandromoreira:
Hey @cdunklau thanks for the contribution, it is amazing such effort!  In regard to your worries, I think this PR doesn't offer any breaking change, it adds features, therefore, all the tests should be passing without any issues.

tests/playlists.py
```diff
@@ -96,6 +96,14 @@
...
 '''
 
+VARIANT_PLAYLIST_WITH_VIDEO_CC_SUBS_AND_AUDIO = '''
+#EXTM3U
+#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=7680000,CLOSED-CAPTIONS="cc",SUBTITLES="sub",AUDIO="aud",VIDEO="vid"
+http://example.com/with-everything-hi.m3u8
+#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=65000,CLOSED-CAPTIONS="cc",SUBTITLES="sub",AUDIO="aud",VIDEO="vid"
+http://example.com/with-everything-low.m3u8
+'''
+
```
leandromoreira: I would use fake uri (i.e: `http://example.com/cc.srt`) instead of `"cc".

```
CLOSED-CAPTION does not specify a Rendition; the closed caption
      media is present in the Media Segments of every video Rendition.

      URI

      The value is a quoted-string containing a URI that identifies the
      Media Playlist file.  This attribute is OPTIONAL; see
      Section 4.3.4.2.1.  If the TYPE is CLOSED-CAPTIONS, the URI
      attribute MUST NOT be present.
```

cdunklau: I think you're conflating the URI *attribute* of EXT-X-I-FRAME-STREAM-INF with the URI *line* that must follow EXT-X-STREAM-INF. In EXT-X-STREAM-INF, the CLOSED-CAPTIONS, AUDIO, VIDEO, and SUBTITLES attributes all reference a "GROUP-ID attribute of an EXT-X-MEDIA tag elsewhere".

However, reviewing this made me realize that StreamInfo for EXT-X-I-FRAME-STREAM-INF should have audio, subtitles, and closed-captions set to None explicitly.

> All attributes defined for the EXT-X-STREAM-INF tag (Section 4.3.4.2)
> are also defined for the EXT-X-I-FRAME-STREAM-INF tag, except for the
> FRAME-RATE, AUDIO, SUBTITLES, and CLOSED-CAPTIONS attributes.

leandromoreira:
Do you plan to do any other PR soon? I'm thinking to release a new version.

PR #129, opened 2018-12-04
Add the ability to set URI's when loading from a string

colde:
Sometimes i want to manage loading of the data myself, but i still would like to have the absolute URI's available on the variant playlists. I can do this manually using the essentially the contents of the _parsed_url function, and then setting the base_uri, this makes it a bit easier without impacting backwards compatibility.


leandromoreira:
Thanks @colde can you provide add a test case for it? 😄

colde:
Test added, is this one fine, or do you want other test cases? There already is a test that verified that my change didn't impact existing cases, which is test_there_should_not_be_absolute_uris_with_loads

Issue #124, opened 2018-09-25
Basic support for custom tags


tausackhn:
Sometimes services add custom tags that they use for their own purposes. Custom tags are also started with `#EXT`. Applications are going to ignore them as it is written in RFC8216. But these tags can be also helpful depending on service. Currently, the library ignores custom tags recognizing them as a comment.
What do you think about saving custom tags started with `#EXT` somewhere in the model? Or specifing additional protocols for `parser` function?

mauricioabreu:
Closed via #130

mauricioabreu:
Hey @tausackhn can you tell us if the change we recently made does what you needed, please?

tausackhn:
@mauricioabreu thank you. It covers my needs.

Issue #123, opened 2018-07-19
Parsing extended #EXTINF tags


hsahovic:
### Context

According to [Wikipedia](https://en.wikipedia.org/wiki/M3U), m3u is a _de facto_ standard, which implies that its implementation might vary across different platforms, software or use cases.

In particular, I have recently been confronted with a couple of implementations that are not compliant with [this spec](https://tools.ietf.org/html/draft-pantos-http-live-streaming-20), which seems to be this repo's reference.

[Here](https://siptv.eu/howto/playlist.html) and [here](https://wiki.tvip.ru/en/m3u) are two simple descriptions of such implementations, and here is an extract of such a m3u8 file :
```
#EXTM3U
#EXTINF:-1 tvg-id="Be1.be" tvg-name="Be1 Alt" group-title="News" parent-code="1234" audio-track="nl",Be1 Name
http://channel.stream.address/to_be_placed/on_separate_line/123.ts
```

### Proposed addition to the parser

**Adding a parser option allowing to parse these extended tags and retrieve them in segments.**

I can fork the project and create a pull request implementing this feature, but I would like to make sure that it fits with the project's orientation.

leandromoreira:
Nice @hsahovic but how do you think to add this extension points? like passing extra functions/classes?  (can you express your ideas with simple example code? like...)
```python
class VariantParser:
    def parse(args):
        return "something"

my_parser = VariantParser()

m3u8_obj = m3u8.load('http://videoserver.com/playlist.m3u8', variant_pipeline=[my_parser])
```

@flavioribeiro is a pythonista and maybe he can help us here too.

hsahovic:
@leandromoreira I was thinking of modifying the `parse` function's signature from:

```python
def parse(content, strict=False)
``` 

To something along the lines of:

```python
def parse(content, strict=False, accept_extended_tags=False)
```

The default behavior of the function would remain the same, but with `accept_extended_tags` set to `True`,  a script such as 

```python
from m3u8 import parse

content = """#EXTM3U
#EXTINF:-1 tvg-id="Be1.be" parent-code="1234",Be1 Name
http://channel.stream.address/to_be_placed/on_separate_line/123.ts"""

# Prints content's first segment
print(parse(content)['segments'][0])
```

[...]

PR #130, opened 2019-01-27
Add basic support for custom tags

mauricioabreu:
Main parser function now receives a function that can be
used to parse the line content and save it into the data store object.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -313,3 +313,16 @@ def test_should_parse_start_with_precise():
 def test_simple_playlist_with_discontinuity_sequence():
     data = m3u8.parse(playlists.SIMPLE_PLAYLIST_WITH_DISCONTINUITY_SEQUENCE)
     assert data['discontinuity_sequence'] == 123
+
+def test_simple_playlist_with_custom_tags():
+    def get_movie(line, data, lineno):
+        custom_tag = line.split(':')
+        if len(custom_tag) == 2:
+            data['movie'] = custom_tag[1].strip()
+
+    data = m3u8.parse(playlists.SIMPLE_PLAYLIST_WITH_CUSTOM_TAGS, strict=False, custom_tags_parser=get_movie)
+    assert data['movie'] == 'million dollar baby'
+    assert 5220 == data['targetduration']
+    assert 0 == data['media_sequence']
+    assert ['http://media.example.com/entire.ts'] == [c['uri'] for c in data['segments']]
+    assert [5220] == [c['duration'] for c in data['segments']]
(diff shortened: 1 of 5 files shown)


mauricioabreu:
I have another idea here: we can add the "ignored" tags in a new data structure, containing the `line`, `lineno` and any other useful information that can be directly used by `m3u8` users after they parse the content.

```python
...
        elif line.startswith('#'):
            data['ignored_lines'] = {'line': line, 'lineno': lineno}
...

# Usage
m3u8_obj = m3u8.loads('#EXTM3U8 ... etc ... ')
data = post_process_data(m3u8_obj.data)
```

tests/test_parser.py
```diff
@@ -313,3 +313,12 @@ def test_should_parse_start_with_precise():
...
     assert data['discontinuity_sequence'] == 123
+
+def test_simple_playlist_with_custom_tags():
+    def get_movie(line, data, lineno):
+        custom_tag = line.split(':')
+        if len(custom_tag) == 2:
+            data['movie'] = custom_tag[1].strip()
+
+    data = m3u8.parse(playlists.SIMPLE_PLAYLIST_WITH_CUSTOM_TAGS, strict=False, custom_tags_parser=get_movie)
+    assert data['movie'] == 'million dollar baby'
```
leandromoreira: Can you assert also the others fields? (so we can make sure that it won't break the existent parser)

mauricioabreu: Done!

leandromoreira:
LGTM and solves https://github.com/globocom/m3u8/issues/123 and https://github.com/globocom/m3u8/issues/124 does all the tests pass locally @mauricioabreu ? I tried to restart the build manually but the same error happened 

```
 Plugin 'pytest_cov' could not be loaded:
```

mauricioabreu:
@leandromoreira I was reading about this error and it seems to be incompatibility of `python version` vs `pytest_cov` plugin. One way to solve this issue is using `tox`: https://tox.readthedocs.io/en/latest/

mauricioabreu:
Python 3.5 and Python 2.6 passed the test suite: https://travis-ci.org/globocom/m3u8/builds/485545837?utm_source=github_status&utm_medium=notification

leandromoreira:
so I'll merge it and then I'll try to use this tox thing.

mauricioabreu:
@leandromoreira we don't pin versions of any dependence. Recently `pytest_cov` deprecated some python versions as we can see here: https://github.com/pytest-dev/pytest-cov/blob/master/setup.py#L127

```
python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*'
```

Maybe `tox` can help us to define different dependency versions for different python versions.

leandromoreira:
We should do this.

leandromoreira:
good catch

PR #132, opened 2019-02-04
Fix conflict of testing requirements

mauricioabreu:
`pytest-cov 2.6.0` has increased the version requirement for the coverage package from `>=3.7.1` to `>=4.4`, which is in conflict with the version requirement defined by the python-coveralls package for `coverage==4.0.3`.

Once `coveralls` upgrades its `coverage` dependency we can get back to a latest pytest-cov version

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,6 @@
 -r requirements.txt
 arrow==0.9.0
-pytest
 bottle
+pytest
 pytest-cov>=2.4.0,<2.6
 python-coveralls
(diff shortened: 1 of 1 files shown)


.venv/.Python
```diff
@@ -0,0 +1 @@
+/System/Library/Frameworks/Python.framework/Versions/2.7/Python
```
leandromoreira: Is this related only for macos?

mauricioabreu: doh! my bad 😞 
I added it unintentionally

PR #134, opened 2019-02-05
Remove unnecessary dependency

mauricioabreu:
datetime from standard library is able to perform the same operations as arrow on m3u8 codebase.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,4 @@
 -r requirements.txt
-arrow==0.9.0
 bottle
 pytest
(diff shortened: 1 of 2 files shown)


leandromoreira:
https://travis-ci.org/globocom/m3u8/builds/488794728?utm_source=github_status&utm_medium=notification

mauricioabreu:
@leandromoreira good catch! I did not remember python 2 did not have explicit support for `UTC`
To compare the datetimes, I had to create a class to represent `UTC`

Issue #109, opened 2017-10-02
#EXT-X-MAP:URI and #EXT-X-PROGRAM-DATE-TIME ignored in dump


mfojtak:
The **#EXT-X-PROGRAM-DATE-TIME** is ignored when dumping m3u8 object to string.

Also the segment level **#EXT-X-PROGRAM-DATE-TIME** is ignored.
The global one gets dumped ok.

mauricioabreu:
Thank you for the effort to open this issue.
Yes, I guess it is a bug. I just opened a pull request to deal with it: https://github.com/globocom/m3u8/issues/109

mauricioabreu:
We will release this bugfix in the next version. Thanks again for contributing to this project.

PR #137, opened 2019-03-10
Missing EXT-X-MAP attributes on dumps

mauricioabreu:


m3u8/model.py
```diff
@@ -263,6 +263,13 @@ def dumps(self):
             output.append(str(self.start))
         if self.is_i_frames_only:
             output.append('#EXT-X-I-FRAMES-ONLY')
+        if self.segment_map:
+            map_output = []
+            for attr, value in self.segment_map.items():
+                if '"' not in value:
```
mauricioabreu: I needed to do it because the `URI` comes without the quotes. They are even defined in the RFC:
https://tools.ietf.org/html/rfc8216#section-4.3.2.5

```
Its format is:

   #EXT-X-MAP:<attribute-list>

   The following attributes are defined:

      URI

      The value is a quoted-string containing a URI that identifies a
      resource that contains the Media Initialization Section.  This
      attribute is REQUIRED.
````

There is a part in the code that removes it. I don't know the reasons behind this but I am pretty sure they exist.

mauricioabreu:
Closes https://github.com/globocom/m3u8/issues/109

Issue #146, opened 2019-07-30
Add support for FRAME-RATE tag


gorzynski:
Tag `FRAME-RATE` is ignored when parsing and dumping a master playlist.

For example when [master-with-stream-inf-1.m3u8](https://github.com/grafov/m3u8/blob/master/sample-playlists/master-with-stream-inf-1.m3u8), which contains `FRAME-RATE` tag, is loaded:
```
m3u8_obj = m3u8.load('master-with-stream-inf-1.m3u8')
```
And then dumped:
```
print(m3u8_obj.dumps())
```
The `FRAME-RATE` tag is gone:
```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=300000,AVERAGE-BANDWIDTH=300000,CODECS="avc1.42c015,mp4a.40.2"
chunklist-b300000.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=600000,AVERAGE-BANDWIDTH=600000,CODECS="avc1.42c015,mp4a.40.2"
chunklist-b600000.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=850000,AVERAGE-BANDWIDTH=850000,CODECS="avc1.42c015,mp4a.40.2"
chunklist-b850000.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=1000000,AVERAGE-BANDWIDTH=1000000,CODECS="avc1.42c015,mp4a.40.2"
chunklist-b1000000.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=1,BANDWIDTH=1500000,AVERAGE-BANDWIDTH=1500000,CODECS="avc1.42c015,mp4a.40.2"
chunklist-b1500000.m3u8
```

The solution would be probably to extend `stream_info` for `Playlist` class.

mauricioabreu:
Good catch! I just opened a new pull request (with a regression test) to fix this bug.
Thank you!

gorzynski:
Actually I already made the change yesterday and forgot to open a pull request. I think it might be a little bit fuller since it can dump FRAME-RATE either it `12.5` or `25` format instead of always adding trailing zeros. Tests added as well. Anyway the pull request is #149 if you feel like using it :).

mauricioabreu:
[...]

PR #148, opened 2019-07-31
Include FRAME-RATE tag on dumps

mauricioabreu:
Fix code to include the tag and add a regression
test to ensure it won't happen in the future.


mauricioabreu:
Closes #146

PR #150, opened 2019-08-08
Don't assume EXTINF titles are quoted

davemevans:
This library currently assumes that EXTINF titles are always single or double quoted when parsing, and always dumps with double quotes.

Reading the RFC, I don't believe this to be the case - the title is simply a `human-readable informative title of the Media Segment expressed as UTF-8 text.`.

Currently `#EXTINF:5220,A sample unquoted title` becomes `#EXTINF:5220,"A sample unquoted title"` when a playlist is parsed then dumped. Similarly `#EXTINF:5220,'A sample single-quoted title'` becomes `#EXTINF:5220,"A sample single-quoted title"`, neither of which are the same as the input.

This patch removes that assumption, fixing the example use cases above. A consequence is that those requiring automatic double quotes on the output would need to modify their use of the library.


mauricioabreu:
You are right. The RFC says nothing about quotes.

> This patch removes that assumption, fixing the example use cases above. A consequence is that those requiring automatic double quotes on the output would need to modify their use of the library.

The only way I see to not break users is adding a conditional on the `m3u8` version. But I don't think it is worth.

PR #153, opened 2019-08-15
Hit 100% test coverage

davemevans:
This PR raises coverage to 100% for all code other than that which is dependent on the version of Python under which the tests are run.


mauricioabreu:
Thank you!!! 🚀

PR #154, opened 2019-08-21
Ensure that final line is terminated as per spec

davemevans:
As per https://tools.ietf.org/html/rfc8216#section-4.1, lines in a Playlist file are terminated with a new line (or a carriage return line feed).

If the manifest ends with an `#EXT-X-ENDLIST` or `#EXT-X-RENDITION-REPORT` tag (or anything other than a list of segments), this is not currently the case.

Usually this doesn't cause any issues in production, but we have found that when the line contains an attribute list ending with an integer (eg `#EXT-X-RENDITION-REPORT:URI="../1M/waitForMSN.php",LAST-MSN=273,LAST-PART=2`), iOS clients will fail to parse the integer.

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -304,6 +304,10 @@ class M3U8(object):
         if self.is_endlist:
             output.append('#EXT-X-ENDLIST')
 
+        # ensure that the last line is terminated correctly
+        if output[-1] and not output[-1].endswith('\n'):
+            output.append('')
+
         return '\n'.join(output)
 
     def dump(self, filename):
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -889,6 +889,13 @@ def test_partial_segment_gap_and_byterange():
 
     assert result == expected
 
+def test_endswith_newline():
+    obj = m3u8.loads(playlists.SIMPLE_PLAYLIST)
+
+    manifest = obj.dumps()
+
+    assert manifest.endswith('\n')
+
 # custom asserts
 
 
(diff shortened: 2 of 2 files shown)


m3u8/model.py
```diff
@@ -304,6 +304,10 @@ def dumps(self):
         if self.is_endlist:
             output.append('#EXT-X-ENDLIST')
 
+        # ensure that the last line is terminated correctly
+        if output[-1] and not output[-1].endswith('\n'):
+            output.append('')
```
mauricioabreu: What is better here?

```python
output.append('')
```

or 

```python
output.append('\r\n')
```

davemevans: It needs to be `''` because of the `('\n').join(output)` that follows. Having a blank line causes the previous unterminated line and a new blank line to be joined using `\n`, giving the required result. Using `\r\n` causes that string to be joined, using `\n`, with the previous line, giving an additional unwanted `\n`.

An example from the test, printing `repr(manifest)`:
`\r\n`: `'#EXTM3U\n#EXT-X-TARGETDURATION:5220\n#EXTINF:5220,\nhttp://media.example.com/entire.ts\n#EXT-X-ENDLIST\n\r\n'` 👎 

`''`: `#EXTM3U\n#EXT-X-TARGETDURATION:5220\n#EXTINF:5220,\nhttp://media.example.com/entire.ts\n#EXT-X-ENDLIST\n` 👍 

I'll update the test to explicitly reject the unwanted result.

Another option I considered would have been to append a newline (eg `output[-1] += '\n'`), but I think the proposed way probably has a minor performance benefit.

mauricioabreu: Oh, good catch! Thank you for the detailed answer.

PR #160, opened 2019-09-13
Make all manifest tags available at top level, not just some

davemevans:
Add (and reorder to match model.py) a bunch of missing tag objects so they are available at the top level like the ones already defined.

It'd be really cool if this could go into the next release to pypi 🤞

### DIFF
--- a/m3u8/__init__.py
+++ b/m3u8/__init__.py
@@ -15,14 +15,16 @@ except ImportError:  # Python 2.x
     from urllib2 import urlopen, Request, HTTPError
     from urlparse import urlparse, urljoin
 
-from m3u8.model import M3U8, Playlist, IFramePlaylist, Media, Segment, PartialSegment, PartialSegmentList, PartInformation, RenditionReport, ServerControl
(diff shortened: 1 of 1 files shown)


m3u8/__init__.py
```diff
@@ -15,14 +15,16 @@
     from urllib2 import urlopen, Request, HTTPError
     from urlparse import urlparse, urljoin
 
-from m3u8.model import M3U8, Playlist, IFramePlaylist, Media, Segment, PartialSegment, PartialSegmentList, PartInformation, RenditionReport, ServerControl
+from m3u8.model import M3U8, Segment, SegmentList, PartialSegment, PartialSegmentList, Key, Playlist, IFramePlaylist, Media, MediaList, PlaylistList, Start, RenditionReport, RenditionReportList, ServerControl, Skip, PartInformation
```
mauricioabreu: This line looks a bit long. Could it be improved using parentheses?

davemevans: Yes, fair point. I'll make the change later.

mauricioabreu: Thank you!!!

davemevans: Fixed via bd764b3

PR #161, opened 2019-09-13
Allow #EXT-X-VERSION to be set as int as well as str

davemevans:
https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-02#section-4.4.1.2 states that #EXT-X-VERSION is an integer.

Currently, attempting to set this tag as an int fails since a string is expected. This PR allows version to be set as an int.

Ideally I'd really like the parser to parse version as an integer since I think it is more intuitive, but this would be a (minor) breaking change since it is currently parsed as a string. What do you think? Particularly since a new minor version is about to be issued, it seems like an ideal time to make this change.

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -269,7 +269,7 @@ class M3U8(object):
         if self.allow_cache:
             output.append('#EXT-X-ALLOW-CACHE:' + self.allow_cache.upper())
         if self.version:
-            output.append('#EXT-X-VERSION:' + self.version)
+            output.append('#EXT-X-VERSION:' + str(self.version))
         if self.target_duration:
             output.append('#EXT-X-TARGETDURATION:' +
                           int_or_float_to_string(self.target_duration))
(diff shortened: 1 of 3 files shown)


mauricioabreu:
I think you meant `patch` right? You added two tests ensuring we can set it as string or integer.

About the code itself: instead of using string concatenation, we could use `format`.

```python
output.append('#EXT-X-VERSION: {0}'.format(self.version))
```

davemevans:
> I think you meant patch right? You added two tests ensuring we can set it as string or integer.

Sorry, I don't understand what you mean. I added both tests to make sure that writing the version was still backwards compatible. I haven't made any change re: parsing as an int as I was looking for advice on whether that was sensible.

> About the code itself: instead of using string concatenation, we could use format.

Yes, that's a good idea. However I wanted to maintain the existing conventions, and string concatenation is currently used everywhere.

mauricioabreu:
@davemevans sorry, I made a mistake. You are correct about the `minor` topic.

About the convention, I think it is a good time to start using `format`. We recently dropped some older Python versions and we intend to support only Python 3 next year.

mauricioabreu:
@davemevans tell me if you will change it. Otherwise I will merge it. 😃

PR #163, opened 2019-09-16
Parse EXT-X-VERSION as an int as per spec

davemevans:
As discussed in #161, `#EXT-X-VERSION` is an `int` in the specification. I think it is more intuitive for version to be parsed as an int rather than a string.

Note that this is a (small) breaking change since an `int` rather than a string will be returned - see the change to `test_should_parse_VERSION` in test_parser.py.

I don't feel super-strongly about this, but it seems more spec-compliant and code using this library looks more sensible (in my opinion)

### DIFF
--- a/m3u8/parser.py
+++ b/m3u8/parser.py
@@ -106,7 +106,7 @@ def parse(content, strict=False, custom_tags_parser=None):
             state['cue_start'] = True
 
         elif line.startswith(protocol.ext_x_version):
-            _parse_simple_parameter(line, data)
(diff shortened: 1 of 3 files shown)


leandromoreira:
>  I think it is more intuitive for version to be parsed as an int rather than a string.

Although I agree with you I don't see anyone using version as a numeric type. We probably won't make any calculations on it but at the same time I don't see any problem to use it as integer and I think it's more compliant with the spec as well.... let's see what the other think about this.

Issue #101, opened 2017-03-28
SSL Error Handling


stoyanovgeorge:
When you parse an https m3u8 manifest file without valid SSL certificate the load function crashes reporting ssl.CertificateError and since the load function is based on URLLIB it would be nice if you could add the load(url, verify = False) statement, to skip the SSL check.

mauricioabreu:
Closed via #165

PR #165, opened 2019-09-27
Option to disable SSL verification

mauricioabreu:
Not every server has SSL enabled with valid
certificates. This commit adds an option to disable
by passing verify_ssl as False when loading a playlist.


mauricioabreu:
Solves #101

PR #166, opened 2019-09-30
Ensure correct dumping of rendition report when LAST-PART is 0

davemevans:
Dumping of LAST-PART was disabled if the value was falsy. This is incorrect since 0 is a valid value for LAST-PART.

I only noticed this now we are actually implementing this stuff 😄

Be good if this could make its way into the version that must be about to drop on pypi?

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -881,7 +881,7 @@ class RenditionReport(BasePathMixin):
         report = []
         report.append('URI=' + quoted(self.uri))
         report.append('LAST-MSN=' + int_or_float_to_string(self.last_msn))
-        if self.last_part:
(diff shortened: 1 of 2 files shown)


davemevans:
I see `0.4.0` already made it on to pypi, which I missed. In which case it'd be really lovely to see a `0.4.1` 😄

mauricioabreu:
@davemevans we will release a version today. Thank you for your contributions.

mauricioabreu:
@davemevans it will be released as `0.5.0` because we introduced backwards incompatibility code.

mauricioabreu:
@davemevans done!

PR #171, opened 2019-10-09
Added Dockerized version of m3u8

krasnova19:


leandromoreira:
thanks @krasnova19 I think maybe a python docker image would be smaller and better for this task, also you probably need to install all the dependencies.

mauricioabreu:
Agreed @leandromoreira 
Also the Python version you choose will be deprecated/unsupported in 2020.
Can you change to a new Python version @krasnova19 ?

PR #184, opened 2019-11-14
SCTE35 CUE-OUT and CUE-IN support

davemevans:
We are using the library to parse playlists containing SCTE markers, and we found a few problems:
1) `#EXT-X-CUE-OUT` with no duration caused an unexpected `ValueError` to be thrown. These are used in VOD streams where the break duration can be calculated easily.
2) `#EXT-X-CUE-OUT` tags are replaced with `#EXT-X-CUE-OUT-CONT` when the playlist is dumped.
3) `#EXT-X-CUE-IN` tags are not in the output when the playlist is dumped.

This PR fixes all three of those issues as well as making some method names more like what they actually do. It also retains backwards compatibility, still passing all the existing tests with no changes.


leandromoreira:
such an amazing work @davemevans thank you very much

m3u8/parser.py
```diff
@@ -325,13 +329,19 @@ def _parse_simple_parameter(line, data, cast_to=str):
...
     res = re.match('.*Duration=(.*),SCTE35=(.*)$', value)
     if res:
         state['current_cue_out_duration'] = res.group(1)
         state['current_cue_out_scte35'] = res.group(2)
 
+def _cueout_no_duration(line):
+    # this needs to be called first since line.split in all other
+    # parsers will throw a ValueError if passed just this tag
+    if line == protocol.ext_x_cue_out:
+        return (None, None)
```
mauricioabreu: Tip: Python does not enforce you to return a tuple. Every return with multiple values will end in a tuple.

m3u8/parser.py
```diff
@@ -348,8 +358,19 @@ def _cueout_envivio(line, state, prevline):
...
+def _cueout_simple(line):
+    # this needs to be called after _cueout_elemental
+    # as it would capture those cues incompletely
+    param, value = line.split(':', 1)
+    res = re.match('^(\d+(?:\.\d)?\d*)$', value)
+    if res:
+        return (None, res.group(1))
+
+def _parse_cueout(line, state, prevline):
+    _cueout_state = (_cueout_no_duration(line)
```
mauricioabreu: `any` function can be used here. Do you agree?

m3u8/parser.py
davemevans: Good knowledge - I learnt something today! However, if ok, I think I will leave it because it matches the existing style of the other cue parsers?

m3u8/parser.py
davemevans: I did consider this, but I don't think so. Because of the way `_cueout_state` is used in the conditional below - it either needs to falsey, or a tuple. `any` can only return a boolean, unless I am missing something?

m3u8/parser.py
mauricioabreu: I am ok with it

m3u8/parser.py
mauricioabreu: You are right!

PR #185, opened 2019-12-06
Allow individual EXT-X-MAP tag for segment

almaslov:
Hi! Current support for EXT-X-MAP is not fully compliant with RFC.
According to https://tools.ietf.org/html/rfc8216#page-17 part 4.3.2.5:

> It applies to every Media Segment that appears after it in the
> Playlist until the next EXT-X-MAP tag or until the end of the
> Playlist.

In my case new EXT-X-MAP tag may appear in playlist due to codec change.
This pull request makes tag segment-specific, instead of playlist-specific.


m3u8/model.py
```diff
@@ -445,6 +442,15 @@ def dumps(self, last_segment):
                 output.append(str(self.key))
                 output.append('\n')
 
+        if last_segment and self.init_section != last_segment.init_section:
+            assert self.init_section, "init section can't be None if previous is not None"
```
mauricioabreu: Not sure if this is the right way to propagate an error.
Users of this feature may need to catch it and handle the lack of a value in `init_section`.
What is your opinion?

m3u8/model.py
```diff
@@ -624,6 +630,47 @@ def __eq__(self, other):
...
+
+    `uri`
+      is a string. ex:: "https://priv.example.com/key.php?r=52"
+
+    `byterange`
+      value of BYTERANGE attribute
+
+    `base_uri`
+      uri the segment comes from in URI hierarchy. ex.: http://example.com/path/to
+
```
mauricioabreu: Can you remove this extra line, please?

m3u8/model.py
```diff
@@ -624,6 +630,47 @@ def __eq__(self, other):
...
+        self.uri = uri
+        self.byterange = byterange
+
+    def __str__(self):
+        output = []
+        if self.uri:
+            output.append('URI=' + quoted(self.uri))
+        if self.byterange:
+            output.append('BYTERANGE=' + self.byterange)
+        return self.tag + ':' + ','.join(output)
```
mauricioabreu: I think you can take advantage of `f strings` here.
```python
return f"{self.tag}:{','.join(output)}"
```

Or even format:

```python
return "{tag}:{attributes}".format(tag=self.tag, attributes=",".join(output))
```

m3u8/model.py
```diff
@@ -624,6 +630,47 @@ def __eq__(self, other):
...
+
+    def __eq__(self, other):
```
mauricioabreu: Nice! 👍

m3u8/model.py
almaslov: Yep! You are right.
This condition should be always true for parsed playlists, but may be violated for user crafted playlist.

m3u8/model.py
almaslov: String interpolation introduced in python3.6, so I think format form is preferred for legacy interpreter's hostages, like me.

m3u8/model.py
almaslov: done

m3u8/model.py
mauricioabreu: You are right!

mauricioabreu:
Thank you @almaslov 
Do you need it to be released soon?

almaslov:
> Do you need it to be released soon?

That would be great

mauricioabreu:
@almaslov done!

Issue #188, opened 2020-01-17
Handling a negative EXT-X-MEDIA-SEQUENCE


wlerin:
So, this is completely nonstandard (decimal-integer is defined to start at 0), but I have encountered a number of streams that provide a negative media-sequence in conjunction with EXT-X-DISCONTINUITY. Thanks to `normalize_attribute`, this library ends up throwing a ValueError when it tries to cast e.g. "_3" to an int. 

Is there any way to work around this? Or can the parser be changed to, well, *not* normalize EXT-X-MEDIA-SEQUENCE values? I'm not entirely sure why it does so currently.

leandromoreira:
a negative media sequence is really not usual maybe you could pass your own parser, I think we build up something that enables its users to do that isn't that right @mauricioabreu ?

mauricioabreu:
@wlerin @leandromoreira 
I think you can use the `custom_tags_parser` feature: https://github.com/globocom/m3u8#custom-tags but I am not 100% sure if it will work.

```python
import m3u8

def media_sequence(line, data, lineno):
    if line.startswith('#EXT-MEDIA-SEQUENCE:'):
        media_sequence = line.split(':')[1]
        data['media_sequence'] = abs(int(media_sequence.strip()))

m3u8_obj = m3u8.load('http://videoserver.com/playlist.m3u8', custom_tags_parser=media_sequence)
print(m3u8_obj.data['media_sequence'])
```

Anyways, I think it could be easier if we just call `abs` inside the m3u8 core. What do you think @leandromoreira ?

wlerin:
```
    m3u8_obj = m3u8.load(src, headers=headers or DEFAULT_HEADERS)
  File "/home/wlerin/.virtualenvs/dev/lib/python3.7/site-packages/m3u8/__init__.py", line 56, in load
    return _load_from_uri(uri, timeout, headers, custom_tags_parser, verify_ssl)
  File "/home/wlerin/.virtualenvs/dev/lib/python3.7/site-packages/m3u8/__init__.py", line 74, in _load_from_uri
    return M3U8(content, base_uri=base_uri, custom_tags_parser=custom_tags_parser)
  File "/home/wlerin/.virtualenvs/dev/lib/python3.7/site-packages/m3u8/model.py", line 147, in __init__
    self.data = parse(content, strict, custom_tags_parser)
  File "/home/wlerin/.virtualenvs/dev/lib/python3.7/site-packages/m3u8/parser.py", line 81, in parse
[...]

PR #189, opened 2020-01-19
Interpret raw values as they are

mauricioabreu:
Before this commit, a negative number was resulting in a bad value. -10 was being replaced with _10, raising undesired exceptions.


mauricioabreu:
Related to #188

wlerin:
So this would affect `ext-x-targetduration`, `ext-x-media-sequence`, `ext-x-discontinuity-sequence`, and `ext-x-version`, all of which are (supposed to be) decimal-integers beginning from 0, and `ext-x-allow-cache` and `ext-x-playlist-type`, which are strings with only a few single-word values allowed. I can't see how this change would break any of them, and it would allow the library to handle non-conforming playlists like the ones I've been encountering better. It might allow nonsensical values to surface, but that's better than a confusing ValueError.

In short, looks good.

PR #193, opened 2020-02-14
Fix invalid escape sequence in regex

davemevans:
`DeprecationWarning: invalid escape sequence \d` was being thrown due to non-raw string being used for regex.

My error, apologies.


mauricioabreu:
Thank you for your contribution! 🤗

PR #196, opened 2020-02-28
Preserve absolute uri when loading a playlist

GerardSoleCa:
# Summary

When the M3U8 library loads the content from a resource uri it will store the absolute uri of the file. We would like to keep the source reference, so we can know the precedence of the playlist at any time.

## Tests
I've added one assert case on a test that was asserting the base_uri. If you think that I should add more tests to the set, just let me know.


mauricioabreu:
Thank you for this contribution @GerardSoleCa 
I'm confused here, sorry. What is the difference between the `uri` used to load the playlist and the `absolute_uri` in your pull request?

PR #198, opened 2020-03-16
Add DATERANGE tag support

davemevans:
This PR adds support for the DATERANGE tag as described at https://tools.ietf.org/html/rfc8216#section-4.3.2.7.

https://developer.apple.com/documentation/http_live_streaming/protocol_extension_for_low-latency_hls_preliminary_specification also allows for DATERANGE tags to be interleaved with partial segments, so that is also implemented.

`X-<client-attribute>` attributes are parsed and dumped (though order may not be maintained), but their values are always considered to be raw strings so, for example, quoted strings will maintain quotes when parsed.

Note that whilst, according to the latest draft, DATERANGE tags are now not technically associated with segments, this implementation does still attach them to segments rather than as a seperate top level modelled entity on M3U8 in order to significantly simplify dumping.


m3u8/model.py
```diff
@@ -1119,6 +1137,57 @@ def dumps(self):
     def __str__(self):
         return self.dumps()
 
+class DateRangeList(TagList):
+    pass
+
+class DateRange(object):
+    def __init__(self, **kwargs):
+        self.id = kwargs.get('id')
```
mauricioabreu: Maybe id and start_date can use the `kwargs['id']` syntax since they are required. What do you think?

m3u8/model.py
```diff
@@ -1119,6 +1137,57 @@ def dumps(self):
...
+        self.scte35_out = kwargs.get('scte35_out')
+        self.scte35_in = kwargs.get('scte35_in')
+        self.end_on_next = kwargs.get('end_on_next')
+        self.x_client_attrs = [ (attr, kwargs.get(attr)) for attr in kwargs if attr.startswith('x_') ]
+
+    def dumps(self):
+        daterange = []
+        daterange.append('ID=' + quoted(self.id))
+
+        if (self.start_date):
```
mauricioabreu: This is required, right? Could not it be written just like the `ID` field above?

m3u8/model.py
```diff
@@ -1119,6 +1137,57 @@ def dumps(self):
...
+        if (self.scte35_out):
+            daterange.append('SCTE35-OUT=' + self.scte35_out)
+        if (self.scte35_in):
+            daterange.append('SCTE35-IN=' + self.scte35_in)
+        if (self.end_on_next):
+            daterange.append('END-ON-NEXT=' + self.end_on_next)
+
+        for attr, value in self.x_client_attrs:
+            daterange.append('%s=%s' % (
+                    denormalize_attribute(attr),
```
mauricioabreu: This indentation does not look right, does it?

m3u8/model.py
davemevans: I thought I had put a comment here to explain why this is so, but I must have forgotten.

The spec says it's required, but there's also an example in the spec that does not contain this (see https://tools.ietf.org/html/rfc8216#section-8.10). Further, we see examples in the wild where this attribute is not used, presumably because of the example mentioned above, so making it required is problematic.

I'll add a comment.

m3u8/model.py
davemevans: It doesn't look right. It seems to work for reasons I don't understand, but i'll correct it anyway.

m3u8/model.py
davemevans: I will change id, but can't change start_date for the reasons described at https://github.com/globocom/m3u8/pull/198#discussion_r393658553

m3u8/model.py
davemevans: Note this case is covered in the provided tests (by using a derivation of the example in the RFC).

m3u8/model.py
davemevans: Fixed via 22bce1e

m3u8/model.py
davemevans: Comment added in 22bce1e.

m3u8/model.py
davemevans: Indentation fixed via 22bce1e.

m3u8/model.py
mauricioabreu: I agree with the comment. Thank you for clarifying this

mauricioabreu:
Thank you @davemevans

mauricioabreu:
can I merge it? @davemevans

davemevans:
Yes please!

PR #199, opened 2020-03-18
Added support for CMSh1Hash param on #EXT-X-KEY tags

isaiah1112:
Ran into these tags while doing key rotation on HLS playlists.  More info on what this value is and does can be found in Section 7.2 of: https://www.adobe.com/content/dam/acom/en/devnet/primetime/PrimetimeHLS_April2014.pdf

For reference:
```
An AAXS EXT-X-KEY tag MAY have the CMSha1Hash attribute. The attribute value is a hexadecimal-integer that contains the message digest that results from applying SHA-1 (as defined by [FIPS.180-2.2002]) to the tag's AAXS DRM metadata. This attribute SHALL be present when the stream makes use of license rotation.
```

Without this you receive the following exception when parsing an HLS Playlist:
```
 File "/Users/anon/bin/m3u8util.py", line 730, in cli_download
   playlist = m3u8.loads(rdata.text)
 File "/Users/anon/.virtualenvs/fd/lib/python3.8/site-packages/m3u8/__init__.py", line 42, in loads
   return M3U8(content, custom_tags_parser=custom_tags_parser)
 File "/Users/anon/.virtualenvs/fd/lib/python3.8/site-packages/m3u8/model.py", line 155, in __init__
   self._initialize_attributes()
 File "/Users/anon/.virtualenvs/fd/lib/python3.8/site-packages/m3u8/model.py", line 160, in _initialize_attributes
   self.keys = [ Key(base_uri=self.base_uri, **params) if params else None
 File "/Users/anon/.virtualenvs/fd/lib/python3.8/site-packages/m3u8/model.py", line 160, in <listcomp>
   self.keys = [ Key(base_uri=self.base_uri, **params) if params else None
TypeError: __init__() got an unexpected keyword argument 'cmsha1hash'
```


leandromoreira:
I think we should fix this `TypeError: __init__() got an unexpected keyword argument 'cmsha1hash'` error but for the `CMSha1Hash` I think you should use the custom parser.

isaiah1112:
I actually did try using the custom parser but I'm not sure if it gets passed to sub-manifests which is where this error is occurring. I'll play with it some more today to see if I can't get working using a custom parser.  If not, perhaps I'll look into submitting a new PR to just handle this error gracefully without adding `CMSha1Hash` to the accepted params.

isaiah1112:
I could do something like this in a different PR if that would be better:
```
-    def __init__(self, method, base_uri, uri=None, iv=None, keyformat=None, keyformatversions=None):
+    def __init__(self, method, base_uri, uri=None, iv=None, keyformat=None, keyformatversions=None, **kwargs):
         self.method = method
         self.uri = uri
         self.iv = iv
         self.keyformat = keyformat
         self.keyformatversions = keyformatversions
         self.base_uri = base_uri
+        self._extra_params = kwargs
```

This would allow the extra parameter to be passed to the Key class but just store them in a dictionary.

isaiah1112:
Closing in favor of: https://github.com/globocom/m3u8/pull/200

PR #201, opened 2020-04-14
Added check that parameter name matches fully

olekhov:
Hello.
I've encountered m3u8 playlist with custom header:
`#EXT-X-MEDIA-READY:7f659f6f09bce196d7`
This header is erroneously parsed as `#EXT-X-MEDIA` header and gives this error:
`ValueError: not enough values to unpack (expected 2, got 1)`
at parser.py: 275

I think you should check that parameter string starts with exact parameter name, not just part of it.
Here is PR that fixes this.
test is also provided.

### DIFF
--- a/m3u8/parser.py
+++ b/m3u8/parser.py
@@ -271,6 +271,9 @@ def _parse_attribute_list(prefix, line, atribute_parser):
     params = ATTRIBUTELISTPATTERN.split(line.replace(prefix + ':', ''))[1::2]
 
     attributes = {}
+    if not line.startswith(prefix+':') :
+        return attributes
+
     for param in params:
         name, value = param.split('=', 1)
         name = normalize_attribute(name)
--- a/tests/playlists.py
+++ b/tests/playlists.py
@@ -86,6 +86,18 @@ http://media.example.com/fileSequence52-2.ts
 http://media.example.com/fileSequence52-3.ts
 '''
 
+PLAYLIST_WITH_CUSTOM_MEDIA_HEADER = '''#EXTM3U
+#EXT-X-VERSION:3
+#EXT-X-TARGETDURATION:6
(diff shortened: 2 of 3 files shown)


tests/playlists.py
```diff
@@ -86,6 +86,18 @@
 http://media.example.com/fileSequence52-3.ts
 '''
 
+PLAYLIST_WITH_CUSTOM_MEDIA_HEADER = '''#EXTM3U
```
mauricioabreu, line 89: I think you can name it `PLAYLIST_WITH_CUSTOM_TAG` because your code fixes every custom tag, not just `EXT-X-MEDIA`. Do you agree?

tests/test_model.py
```diff
@@ -273,6 +273,10 @@ def test_session_key_attribute_without_initialization_vector():
     assert None == obj.session_keys[0].iv
 
 
+def test_parse_custom_media_header():
+    obj = m3u8.M3U8(playlists.PLAYLIST_WITH_CUSTOM_MEDIA_HEADER)
```
mauricioabreu, line 277: I think we should expect this test to assert something. I get it does not break, but what happens with the EXT-X-MEDIA-READY? Does it go away? Does it get parsed?

m3u8/parser.py
```diff
@@ -271,6 +271,9 @@ def _parse_attribute_list(prefix, line, atribute_parser):
     params = ATTRIBUTELISTPATTERN.split(line.replace(prefix + ':', ''))[1::2]
 
     attributes = {}
+    if not line.startswith(prefix+':') :
```
mauricioabreu, line 274: Nice catch! Could you remove that extra black space before the last colon?

mauricioabreu:
Thank you for your contribution @olekhov 
I merged it into the master branch with a few fixes.

PR #202, opened 2020-04-19
Add more rendition support to StreamInfo.

Mz49:
Pages [23](https://tools.ietf.org/html/draft-pantos-http-live-streaming-16#page-23) and [24](https://tools.ietf.org/html/draft-pantos-http-live-streaming-16#page-23) of draft-pantos-http-live-streaming-16 specify that stream info may contain AUDIO, VIDEO and SUBTITLES tags which specify the renditions that can be used for the media. This pull request adds the ability for StreamInfo to also dump it's `audio`, `video` and `subtitles` attributes.

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -898,6 +898,12 @@ class StreamInfo(object):
             stream_inf.append('FRAME-RATE=%.5g' % self.frame_rate)
         if self.codecs is not None:
             stream_inf.append('CODECS=' + quoted(self.codecs))
+        if self.audio is not None:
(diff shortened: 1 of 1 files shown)


leandromoreira:
Isn't that already done via https://github.com/globocom/m3u8/blob/6c329a2c473ab59a873e17414262fefa4888498f/m3u8/parser.py#L288 ?

Mz49:
I apologize for my late reply. It does parse it, but it does not dump it. My pull request attempts to fix that.

mauricioabreu:
@Mz49 thanks.
Your tests are failing. Can you check your code again?

PR #203, opened 2020-04-20
EXT-X-GAP support

davemevans:
As defined https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-05#section-4.4.2.7 and in the draft LLHLS extensions.

Crafted during these long COVID nights ...

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,6 +64,7 @@ Supported tags
 * #EXT-X-SKIP
 * `#EXT-X-SESSION-DATA`_
 * `#EXT-X-DATERANGE`_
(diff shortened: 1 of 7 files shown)


davemevans:
Hmm, the test failure relates to `test_daterange_simple` which hasn't changed here.
That test passed with the same test setup in eg https://travis-ci.org/github/globocom/m3u8/jobs/675005951 (via #201) 😕

mauricioabreu:
@davemevans not sure what is going on with the tests.
I will merge it anyways. Thank you for all your hard work on this project.

PR #209, opened 2020-05-15
Added support for EXTTV and keeping of custom tags

bokysan:
This commit:
- Adds support for `EXTV`: I haven't found any official specification on
  the format, but seems to be used throughout the web.
- Add support for keeping custom tags on the lists. This means it's now
  possible to do a `load()` / `dumps()` cycle without loosing any
  metadata.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -84,6 +84,8 @@ Supported tags
 * `#EXT-X-SESSION-DATA`_
 * `#EXT-X-DATERANGE`_
 * `#EXT-X-GAP`_
+* `#EXTTV`
+* extra tags
 
 Encryption keys
 ---------------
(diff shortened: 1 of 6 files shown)


mauricioabreu:
Hello @bokysan thank you for the contribution.
What does this `EXTTV` do?

bokysan:
Here's info [from Google](https://www.google.com/search?q=%22%23EXTTV).

Basically additional info for IPTV channels. Namely, the structure seems to be as follows:
```
#EXTTV:Tag1-Tag2-Tag3;lang;XML-TV-Channel;Channel_Icon
```

So:
- a list of tags attached to the channel, separated by commads (e.g. "movies,fun,witty")
- language of the channel
- XML EPG id
- channel icon

leandromoreira:
@mauricioabreu isn't the case of custom parser usage instead of code here?

mauricioabreu:
@leandromoreira I am not sure if it will work. I will try to achieve the same goal of this PR by using the custom parser.

PR #212, opened 2020-06-11
Remove Python2 references

mauricioabreu:
Python2 is no longer supported.


leandromoreira:
I'd like to know what @igorsobreira @flavioribeiro or @hltbra think about it, I'm in favor.

PR #213, opened 2020-06-11
Allow users to use custom HTTP clients

mauricioabreu:
With this change, users can now use `requests` or another HTTP client.
Besides that, users can now use proxies by passing `proxies` to the default client constructor.

### DIFF
--- a/m3u8/httpclient.py
+++ b/m3u8/httpclient.py
@@ -1,0 +1,58 @@
+import posixpath
+import ssl
+import sys
+
+try:
+    import urllib
+    from urllib.error import HTTPError
+    from urllib.parse import urlparse, urljoin
+except ImportError:  # Python 2.x
+    from urllib2 import urlopen, Request, HTTPError
+    from urlparse import urlparse, urljoin
+
+
+PYTHON_MAJOR_VERSION = sys.version_info
+
+
+def _parsed_url(url):
+    parsed_url = urlparse(url)
+    prefix = parsed_url.scheme + '://' + parsed_url.netloc
+    base_path = posixpath.normpath(parsed_url.path + '/..')
+    return urljoin(prefix, base_path)
+
+def _read_python2x(resource):
+    return resource.read().strip()
+
+
+def _read_python3x(resource):
+    return resource.read().decode(
+        resource.headers.get_content_charset(failobj="utf-8")
+    )
+
+
+class DefaultHTTPClient:
+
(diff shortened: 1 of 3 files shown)


m3u8/httpclient.py
```diff
@@ -0,0 +1,59 @@
+import posixpath
+import ssl
+import sys
+
+try:
+    import urllib
+    from urllib.error import HTTPError
+    from urllib.parse import urlparse, urljoin
+except ImportError:  # Python 2.x
```
rokam, line 9: For Puthon 2.x support you need to add :
`from urllib.request import Request`

m3u8/httpclient.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+
+
+class DefaultHTTPClient:
+
+    def __init__(self, proxies=None):
+        self.proxies = proxies
+
+    def download(self, uri, timeout=None, headers={}, verify_ssl=True):
+        proxy_handler = urllib.request.ProxyHandler(self.proxies)
```
rokam, line 39: Change urllib.request for Request

m3u8/httpclient.py
```diff
@@ -0,0 +1,59 @@
...
+        https_handler = HTTPSHandler(verify_ssl=verify_ssl)
+        opener = urllib.request.build_opener(proxy_handler, https_handler)
```
rokam, line 41: Change urllib.request for Request

mauricioabreu:
@rokam Thank you for the review. We want to drop support for Python 2 since it is no longer supported by Python Software Foundation.

rokam:
@mauricioabreu that way, I do think you should remove the python2 import from httpclient.py. And also, remove the 2.7 version from .travis.yml.

mauricioabreu:
@rokam done, thank you!

leandromoreira:
are we going to remove py2.7 in travis-ci here in this PR ?

mauricioabreu:
@leandromoreira I would like to keep the "drop Python 2" feature in the #212 so we have 2 separate git histories for two different problems. Looks good?

mauricioabreu:
@leandromoreira but first we need to accept and merge the #212

mauricioabreu:
@leandromoreira but first we need to accept and merge the #212

mauricioabreu:
Thank you @leandromoreira and @rokam

Issue #206, opened 2020-04-26
about #EXTINF


liuyug:
I have checked "HTTP Live Streaming" protocol that is said the format is ```#EXTINF:<duration>,[<title>]```. Your library follow it. 

But I found many M3U instance which format is ```#EXTINF:-1 tvg-id="" ....``` from https://github.com/iptv-org/iptv

May you give me advice how to load and process them?

SerhiyRomanov:
I'm faced with the same issue - I am trying to parse a playlist from my IPTV provider and it contains lines like this 
```
#EXTINF:-1 timeshift="7" catchup-days="7" catchup-type="flussonic-ts" tvg-id="first"  group-title="Common folder" tvg-logo="http://pl.myiptv.tv:80/icon/9195.png",First Channel
http://str04.myiptv.tv/9195/mpegts?username=user&password=password&token=token&ch_id=70&req_host=pkjX3BL
```

I don't find any words about this tags here https://tools.ietf.org/html/rfc8216#section-4.3.2.1

Implementation of parsing these tags seems not very complicated, but would it be acceptable?

leandromoreira:
I think you can use https://github.com/globocom/m3u8#custom-tags

liuyug:
got it. Thank you

PR #214, opened 2020-07-02
WIP: Added parsing the additional props related to IPTV playlists

SerhiyRomanov:
In this PR was added the ability to work with additional properties in EXTINF tag, which are widely used with IPTV playlists.
It's related to https://github.com/globocom/m3u8/issues/123 and https://github.com/globocom/m3u8/issues/206

For instance, my IPTV provider generates a playlist like this:
```
#EXTM3U url-tvg="http://ottg.tv/epg.xml.gz"
#EXTINF:-1 timeshift="7" catchup-days="7" catchup-type="flussonic-ts" tvg-id="pervy"  group-title="News" tvg-logo="http://pl.ottg.tv:80/icon/9195.png",First Channel
http://str04.ottg.tv/9195/mpegts?username=user&password=pass&token=token&ch_id=70&req_host=pkjX3BL
#EXTINF:-1 timeshift="7" catchup-days="7" catchup-type="flussonic-ts" tvg-id="pervy"  group-title="Films" tvg-logo="http://pl.ottg.tv:80/icon/2214.png",First Channel HD
http://str04.ottg.tv/2214/mpegts?username=user&password=pass&token=token&ch_id=76&req_host=pkjX3BL
```

I've not found any official specifications about such properties, except for a mention on Wiki https://en.wikipedia.org/wiki/M3U and some explanations here https://en.f-player.ru/extm3u-iptv-format
But from my experience, many IPTV-players are reading and use such properties, for instance, Kodi (XMBC) https://github.com/kodi-pvr/pvr.iptvsimple#m3u-format-elements

I understand, that the same functionality I can get with using custom parser for tags, but it seems, that would be better to have such ability built-in.

### Features that have not still implemented
1. Parsing additional props from  #EXTM3U:
  `#EXTM3U url-tvg="http://ottg.tv/epg.xml.gz"`
_url-tvg_ property says to player which EPG source use
2. Parsing tags: _#PLAYLIST:_

I would like to discuss the implementation. 
In the first case, it seems like we can add a property additional_props (type: dict) to M3U8 class.

And I have no idea what to do with #PLAYLIST tag.

I would be happy to hear any comments about all this stuff.


rokam:
IDK if this should be merged because it's not part of the IETF standard.

SerhiyRomanov:
Yes, I know that this is not standard. But I need this changes for my project, so my fork will stay alive in any case. 
I plan to do update from current master on my fork, if this changes won't be merged.

Issue #216, opened 2020-07-09
frame rate should be rounded to 3 decimal places


thenewguy:
Per https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-07#page-41

Should probably use `'FRAME-RATE=%.3g' % self.frame_rate` instead of `'FRAME-RATE=%.5g' % self.frame_rate` at https://github.com/globocom/m3u8/blob/master/m3u8/model.py#L915

leandromoreira:
Agreed.

mauricioabreu:
Thank you @thenewguy

PR #217, opened 2020-07-09
Rounded framerate to 3 decimal places per spec

thenewguy:
Fixes issue #216


leandromoreira:
hey @thenewguy can you add some tests to ensure this behavior?

* a integer framerate
* a double digits float framerate
* a 5 digits float framerate

tell me if you need any assistance.

mauricioabreu:
Agree with @leandromoreira 

Nice job anyways @thenewguy

thenewguy:
> hey @thenewguy can you add some tests to ensure this behavior?
> 
> * a integer framerate
> * a double digits float framerate
> * a 5 digits float framerate
> 
> tell me if you need any assistance.

Integer and float tested already at - VARIANT_PLAYLIST_WITH_FRAME_RATE

I will add a test that reads a long 5 digit one in and strings it back to 3 digits

tests/test_parser.py
```diff
@@ -400,6 +400,11 @@ def test_master_playlist_with_frame_rate():
     assert 60 == playlists_list[2]['stream_info']['frame_rate']
     assert 12.5 == playlists_list[3]['stream_info']['frame_rate']
 
+def test_master_playlist_with_unrounded_frame_rate():
+    data = m3u8.parse(playlists.VARIANT_PLAYLIST_WITH_ROUNDABLE_FRAME_RATE)
+    playlists_list = list(data['playlists'])
+    assert 12.54321 == playlists_list[0]['stream_info']['frame_rate']
```
leandromoreira, line 406: Shouldn't be already fixed 3 digits here?

```python
     assert 12.543 == playlists_list[0]['stream_info']['frame_rate'] 
```

thenewguy: This verifies we can parse a playlist containing a frame rate longer than 3 digits - it isn't rounded until it is converted back to string with dumps

PR #224, opened 2020-08-02
Fixed file loading decode issue in non-english platform

frank89722:
Just a small fix for local file loading issue in non-english platform


mauricioabreu:
Thank you @frank89722

PR #226, opened 2020-09-04
Added additional parsing for simple CUE-OUT with Duration key

ahardwick:
The following was not parsed by any of the existing CUE-OUT lines

#EXT-X-CUE-OUT:DURATION=120

I added a new parse rather than modifying the existing regex for _cueout_simple


tests/playlists.py
```diff
@@ -776,6 +776,18 @@
 2.aac
 '''
 
+CUE_OUT_WITH_DURATION_KEY_PLAYLIST = '''#EXTM3U
+#EXT-X-TARGETDURATION:6
+#EXT-X-CUE-OUT:DURATION=11.52
```
leandromoreira, line 781: I think there is also a case for `#EXT-X-CUE-OUT:30.000` https://docs.aws.amazon.com/mediatailor/latest/ug/hls-ad-markers.html#hls-ad-markers-cue

leandromoreira (changes requested):

m3u8/parser.py
```diff
@@ -375,6 +375,15 @@ def _cueout_envivio(line, state, prevline):
     else:
         return None
 
+def _cueout_duration(line):
```
davemevans, line 378: This doesn't seem to get called - you need to add it to `_parse_cueout`

tests/playlists.py
```diff
@@ -776,6 +776,18 @@
 2.aac
 '''
 
+CUE_OUT_WITH_DURATION_KEY_PLAYLIST = '''#EXTM3U
```
davemevans, line 779: This doesn't appear to get used in a test anywhere ...

tests/playlists.py
davemevans: That case should be handled by existing code (`_cueout_simple`), assuming this PR doesn't change that.

ahardwick:
I'll updated "_parse_cueout()". Apologies, I meant to add that.

tests/playlists.py
ahardwick: It is handled by the existing cue-out simple

tests/playlists.py
ahardwick: Updated parser.py "_parse_cueout"

m3u8/parser.py
ahardwick: updated parser.py "_parse_cueout"

tests/playlists.py
ahardwick: I didn't include a test as it didn't seem needed, similar to _cueout_simple.  I included it for reference on the notation.

leandromoreira (approved):

mauricioabreu (approved):
Good job and thank you for adding the test with duration.

PR #231, opened 2020-10-10
Adds AVERAGE-BANDWIDTH to IFramePlaylist (resolves #221)

### LINKED ISSUE #221: Output Average Bandwidth for IFramePlaylist
iframe_stream_info sets average bandwidth to None at https://github.com/globocom/m3u8/blob/master/m3u8/model.py#L847

It should be passed along if available per https://tools.ietf.org/html/rfc8216#section-4.3.4.3 and https://tools.ietf.org/html/rfc8216#section-4.3.4.2

RafalLukawiecki:
Adds AVERAGE-BANDWIDTH to IFramePlaylist class to comply with RFC8216 sections 4.3.4.2 and 4.3.4.3.

Resolves #221.


m3u8/model.py
```diff
@@ -861,6 +861,9 @@ def __str__(self):
         if self.iframe_stream_info.bandwidth:
             iframe_stream_inf.append('BANDWIDTH=%d' %
                                      self.iframe_stream_info.bandwidth)
+        if self.iframe_stream_info.average_bandwidth:
+            iframe_stream_inf.append('AVERAGE-BANDWIDTH=%d' %
```
mauricioabreu, line 865: Can you please add a test for this code? I can see a typo here `iframe_stream_inf`. It should be `iframe_stream_info`, right?

mauricioabreu (changes requested):

m3u8/model.py
RafalLukawiecki: Thank you @mauricioabreu for your feedback. Interestingly, if `iframe_stream_inf` is a typo, then it is one I have copied from line 862. The pre-existing code also referred to `iframe_stream_inf` and not to `iframe_stream_info`. I will have a look at it later, just in case the typo is the official way forward.

Happy to add a test.

RafalLukawiecki: @mauricioabreu, I had a look at git blame going back 9 years and it seems to me that `iframe_stream_inf` is how it has been spelled ever since, meaning that the code above has always referred to 'self.iframe_stream_info.xyz` yet the `append` is being applied to `iframe_stream_inf`.

Do you think the remainder of the code that has, I assume, been in use for a while, needs changing? I wonder come it has not been noticed earlier.

mauricioabreu: You are right @RafalLukawiecki sorry!

Could you please add a test then? Thank you for your contribution! (after the test is added I will merge it)

RafalLukawiecki: I have added a couple of tests, both to the parser and to the generator. As a result, I have also added the correct type information for the average_bandwidth attribute to `parser.py`. Hopefully this helps, let me know if you have any other suggestions.

mauricioabreu (approved):
Awesome job! Your tests are very well written. :+1:

leandromoreira (approved):

leandromoreira:
thanks @RafalLukawiecki , @mauricioabreu when wer are going to release a new version let's add the proper tag on the read.me https://github.com/globocom/m3u8#supported-tags

RafalLukawiecki:
> thanks @RafalLukawiecki , @mauricioabreu when wer are going to release a new version let's add the proper tag on the read.me https://github.com/globocom/m3u8#supported-tags

You are very welcome @leandromoreira and @mauricioabreu. As for adding info on the tag in the readme, you already show the `#EXT-X-I-FRAMES-ONLY` tag there. This PR merely adds support for an _attribute_ of that tag. As far as I can see you are not showing a majority of tag attributes in the readme—there are a lot of them, otherwise, to add. :)

mauricioabreu:
Thank you @RafalLukawiecki I will review these tags later in the day and release a new version containing your feature.

PR #236, opened 2020-11-19
Adds VIDEO-RANGE to IFramePlaylist

hjmallon:
VIDEO-RANGE is also useful for IFramePlaylists.

# VIDEO-RANGE

https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-07#section-4.4.6.2

      VIDEO-RANGE

      The value is an enumerated-string; valid strings are SDR, HLG and
      PQ.

      The value MUST be SDR if the video in the Variant Stream is
      encoded using one of the following reference opto-electronic
      transfer characteristic functions specified by the
      TransferCharacteristics code point: [CICP] 1, 6, 13, 14, 15.  Note
      that different TransferCharacteristics code points can use the
      same transfer function.

      The value MUST be HLG if the video in the Variant Stream is
      encoded using a reference opto-electronic transfer characteristic
      function specified by the TransferCharacteristics code point 18,
      or consists of such video mixed with video qualifying as SDR (see
      above).

      The value MUST be PQ if the video in the Variant Stream is encoded
      using a reference opto-electronic transfer characteristic function
      specified by the TransferCharacteristics code point 16, or
      consists of such video mixed with video qualifying as SDR or HLG
      (see above).

      This attribute is OPTIONAL.  Its absence implies a value of SDR.
      Clients that do not recognize the attribute value SHOULD NOT
      select the Variant Stream.

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -820,7 +820,7 @@ class IFramePlaylist(BasePathMixin):
     Attributes:
 
     `iframe_stream_info` is a named tuple containing the attributes:
-     `program_id`, `bandwidth`, `average_bandwidth`, `codecs` and
+     `program_id`, `bandwidth`, `average_bandwidth`, `codecs`, `video_range` and
     `resolution` which is a tuple (w, h) of integers
 
     More info: http://tools.ietf.org/html/draft-pantos-http-live-streaming-07#section-3.3.13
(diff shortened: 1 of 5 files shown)


rogergodspeed:
This is awesome! Good catch, @hjmallon!

leandromoreira (approved):

mauricioabreu (approved):
Good job!

PR #237, opened 2020-11-29
Fix .gitignore - remove .python-version and .idea/

dekoza:
`.python-version` should not be included in VCS. While at it, I also removed `.idea` (PyCharm settings dir) just to be on the safe side.


mauricioabreu:
Thank you @dekoza you are right!

PR #246, opened 2021-03-24
Fix URLs join failed when path contains `//`

### LINKED ISSUE #245: urllib.parse.urljoin use is wrong in case double slash path
Now m3u8
```python
>>> import os
>>> import posixpath
>>> from urllib.parse import urlparse, urljoin
>>>
>>> def _parsed_url(url):
...     parsed_url = urlparse(url)
...     prefix = parsed_url.scheme + '://' + parsed_url.netloc
...     base_path = posixpath.normpath(parsed_url.path + '/..')
...     return urljoin(prefix, base_path)
...
>>> def _urijoin(base_uri, path):
...     if base_uri.startswith('http'):
...         return urljoin(base_uri, path)
...     else:
...         return os.path.normpath(os.path.join(base_uri, path.strip('/')))
...
>>> # above codes just for view, here import real functoins
>>> from m3u8 import _parsed_url
>>> from m3u8.mixins import _urijoin
>>>
>>> # _parsed_url
>>> _parsed_url('http://hostname/path1/path2')  # normal path, only this is right
'http://hostname/path1'
>>> _parsed_url('http://hostname//path1/path2')  # double slash in front of the path
'http://path1'
>>> _parsed_url('http://hostname/path1//path2')  # double slash in middle of the path
'http://hostname/path1'
>>>
>>> # _urijoin
>>> _urijoin('http://hostname/path1/', 'path2')  # normal path, only this is right
'http://hostname/path1/path2'
>>> _urijoin('http://hostname//path1/', 'path2')  # double slash in front of the path
'http://hostname/path1/path2'
>>> _urijoin('http://hostname/path1//', 'path2')  # double slash in end of the path
'http://hostname/path1/path2'
>>> _urijoin('http://hostname/path1/', '//path2')  # double slash in front of the second path
'http://path2'
```
Changes
```python
>>> import os
>>> from urllib.parse import SplitResult, urlsplit
>>>
>>> def _parsed_url(url):
...     parsed_url = urlsplit(url)
...     base_path = parsed_url.path.rpartition('/')[0] + '/'
...     return SplitResult(parsed_url.scheme, parsed_url.netloc, base_path, '', '').geturl()
...
>>> def _urijoin(base_uri, path):
...     if path[0] == '/':
...         path = path[1:]
...     if base_uri.startswith('http'):
...         return base_uri + path
...     else:
...         return os.path.normpath(os.path.join(base_uri, path))
...
>>> # _parsed_url
>>> _parsed_url('http://hostname/path1/path2')  # normal path
'http://hostname/path1/'
>>> _parsed_url('http://hostname//path1/path2')  # double slash in front of the path
'http://hostname//path1/'
>>> _parsed_url('http://hostname/path1//path2')  # double slash in middle of the path
'http://hostname/path1//'
>>>
>>> # _urijoin
>>> _urijoin('http://hostname/path1/', 'path2')  # normal path
'http://hostname/path1/path2'
>>> _urijoin('http://hostname//path1/', 'path2')  # double slash in front of the path
'http://hostname//path1/path2'
>>> _urijoin('http://hostname/path1//', 'path2')  # double slash in end of the path
'http://hostname/path1//path2'
>>> _urijoin('http://hostname/path1/', '//path2')  # double slash in front of the second path
'http://hostname/path1//path2'
```

SeaHOH:
close #245

### DIFF
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -74,13 +74,13 @@ def test_load_should_create_object_from_uri_with_relative_segments():
     prefix = urlparsed.scheme + '://' + urlparsed.netloc
     expected_key_abspath = '%s%s/key.bin' % (prefix, os.path.normpath(base_uri + '/..'))
     expected_key_path = '../key.bin'
-    expected_ts1_abspath = '%s/entire1.ts' % (prefix)
+    expected_ts1_abspath = '%s%s/entire1.ts' % (prefix, base_uri)
     expected_ts1_path = '/entire1.ts'
-    expected_ts2_abspath = '%s%sentire2.ts' % (prefix, os.path.normpath(base_uri + '/..') + '/')
+    expected_ts2_abspath = '%s%s/entire2.ts' % (prefix, os.path.normpath(base_uri + '/..'))
     expected_ts2_path = '../entire2.ts'
     expected_ts3_abspath = '%s%sentire3.ts' % (prefix, os.path.normpath(base_uri + '/../..'))
     expected_ts3_path = '../../entire3.ts'
-    expected_ts4_abspath = '%s%sentire4.ts' % (prefix, base_uri + '/')
+    expected_ts4_abspath = '%s%s/entire4.ts' % (prefix, base_uri)
     expected_ts4_path = 'entire4.ts'
 
     assert isinstance(obj, m3u8.M3U8)
--- a/m3u8/httpclient.py
+++ b/m3u8/httpclient.py
@@ -3,15 +3,14 @@ import ssl
 import sys
 import urllib
 from urllib.error import HTTPError
-from urllib.parse import urlparse, urljoin
+from urllib.parse import SplitResult, urlsplit
 import urllib.request
 
 
 def _parsed_url(url):
-    parsed_url = urlparse(url)
-    prefix = parsed_url.scheme + '://' + parsed_url.netloc
-    base_path = posixpath.normpath(parsed_url.path + '/..')
-    return urljoin(prefix, base_path)
+    parsed_url = urlsplit(url)
+    base_path = parsed_url.path.rpartition('/')[0] + '/'
+    return SplitResult(parsed_url.scheme, parsed_url.netloc, base_path, '', '').geturl()
 
 
 class DefaultHTTPClient:
--- a/m3u8/httpclient.py
+++ b/m3u8/httpclient.py
@@ -1,4 +1,3 @@
-import posixpath
 import ssl
 import sys
 import urllib
--- a/m3u8/mixins.py
+++ b/m3u8/mixins.py
@@ -7,6 +7,8 @@ def _urijoin(base_uri, path):
     if path[0] == '/':
         path = path[1:]
     if base_uri.startswith('http'):
+        if base_uri[-1] != '/':
+            base_uri += '/'
         return base_uri + path
     else:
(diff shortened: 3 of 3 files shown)


SeaHOH:
@leandromoreira @mauricioabreu
Nobody gives a reveiw?

mauricioabreu (approved):

mauricioabreu:
@SeaHOH does this happen when `/` is duplicated without any reason, right?
If this is the problem, maybe replacing `//` by `/` would be enough. What do you think?

SeaHOH:
> does this happen when `/` is duplicated without any reason, right?

I had meet it a m3u8 url which contains `//`, so #245 has been submitted.

> If this is the problem, maybe replacing `//` by `/` would be enough. What do you think?

No, that could not resolve problem.

tests/test_loader.py
```diff
@@ -74,13 +74,13 @@ def test_load_should_create_object_from_uri_with_relative_segments():
     prefix = urlparsed.scheme + '://' + urlparsed.netloc
     expected_key_abspath = '%s%s/key.bin' % (prefix, os.path.normpath(base_uri + '/..'))
     expected_key_path = '../key.bin'
-    expected_ts1_abspath = '%s/entire1.ts' % (prefix)
+    expected_ts1_abspath = '%s%s/entire1.ts' % (prefix, base_uri)
```
leandromoreira, line 77: if what we want to treat is when ta malformed manifest is introducing a double `//` I think we should create a new entire test, do you agree?

leandromoreira: or if we want to test legit URLs with `//` I think we should create a test demonstrating that, what do you think?

leandromoreira (commented):
> > does this happen when `/` is duplicated without any reason, right?
> 
> I had meet it a m3u8 url which contains `//`, so #245 has been submitted.
> 
> > If this is the problem, maybe replacing `//` by `/` would be enough. What do you think?
> 
> No, that could not resolve problem.

Do you think that if we treat `http://hostname//path1/path2` as `http://hostname/path1/path2` wouldn't be better?

tests/test_loader.py
SeaHOH: This is just a fix for new function result.

Of course new testcases may be needed, but notice that double means two or more if you add new.

SeaHOH:
> Do you think that if we treat `http://hostname//path1/path2` as `http://hostname/path1/path2` wouldn't be better? or do you need to keep the double `//` ?

Yes, we need keep it as double or more.

tests/test_loader.py
futzu: **Doesn't urljoin give you what you want?**

```python3
>>>> from urllib.parse import urljoin

>>>> baseurl="https://futzu.com//"
>>>> urljoin(baseurl,"/seg.ts")

'https://futzu.com/seg.ts'

>>>> baseurl="//"
>>>> urljoin(baseurl,"/seg.ts")

'/seg.ts'

>>>> urljoin(baseurl,"//seg.ts")

'//seg.ts'

>>>> urljoin(baseurl,"//seg.ts//")

'//seg.ts//'
```

SeaHOH: > Doesn't urljoin give you what you want?

It doesn't.
```python
>>> from urllib.parse import urljoin
>>> baseurl = "https://futzu.com/"
>>> urljoin(baseurl, "//seg.ts")
'https://seg.ts'
```

PR #249, opened 2021-05-25
Add support for webvtt syncing with timestamp map.

papaver:
Support for webvtt subtitles.  Adds the `USP-X-TIMESTAMP-MAP` tag.  Allows both reading and writing of the tag.  Follows the program timestamp pattern.

```
#EXTM3U
#EXT-X-INDEPENDENT-SEGMENTS
#EXT-X-MEDIA-SEQUENCE:45457
#EXT-X-VERSION:4
#EXT-X-PROGRAM-DATE-TIME:2021-05-24T23:58:46.170000
#USP-X-TIMESTAMP-MAP:MPEGTS=726546,LOCAL=2021-05-24T23:58:46.170000
#EXTINF:5,
/movie/55/subtitle/1/segment/1.vtt
#EXT-X-DISCONTINUITY
#EXT-X-PROGRAM-DATE-TIME:2021-05-24T23:58:51.170000
#USP-X-TIMESTAMP-MAP:MPEGTS=0,LOCAL=2021-05-24T23:58:51.170000
#EXTINF:6,
/movie/64/subtitle/1/segment/0.vtt
#EXTINF:5,
/movie/64/subtitle/1/segment/1.vtt
```


leandromoreira:
Hey @papaver thanks for the PR however we don't support unofficial tags, but you can add them via [custom tag extensions](https://github.com/globocom/m3u8#custom-tags).

davemevans:
Subtitle sync is maintained by the [X-TIMESTAMP-MAP](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-09#section-3.1.4) tag in the VTT segments themselves rather than an entry in the VTT playlist.

papaver:
@leandromoreira i didn't see anywhere where custom tags can be written back to the m3u8.  am i missing something?

papaver:
@davemevans i see. so an entry in the playlist is moot as long as the webvtt file has the correct X-TIMESTAMP-MAP setup in it?

leandromoreira:
> @leandromoreira i didn't see anywhere where custom tags can be written back to the m3u8. am i missing something?

I really don't know, I'm sure you can pass custom tag during load but not sure about the write... @mauricioabreu any thoughts?

PR #258, opened 2021-08-29
Fix bug: media_sequence can be zero

mojtaba-arvin:
The zero media sequence was ignored in the media playlist. (The zero media sequence as default is important for the proper functioning of the players.)

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -288,7 +288,7 @@ class M3U8(object):
         output = ['#EXTM3U']
         if self.is_independent_segments:
             output.append('#EXT-X-INDEPENDENT-SEGMENTS')
-        if self.media_sequence:
+        if not self.is_variant and self.media_sequence is not None:
             output.append('#EXT-X-MEDIA-SEQUENCE:' + str(self.media_sequence))
         if self.discontinuity_sequence:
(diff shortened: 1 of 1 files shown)


davemevans:
This breaks a bunch of the existing tests, and has no accompanying test.

Also, the spec is explicit that if the tag is not in the playlist the value shall be considered to be 0 so it's interesting that this breaks players?

mojtaba-arvin:
> This breaks a bunch of the existing tests, and has no accompanying test.
> 
> Also, the spec is explicit that if the tag is not in the playlist the value shall be considered to be 0 so it's interesting that this breaks players?

    
     The Seek bar of the player will break and the player will show only ten seconds (the length of a segment) of a one-minute HLS. 
     
     Tested on this playlist :
     https://github.com/globocom/m3u8/pull/259#issuecomment-910399500

mauricioabreu:
I'm closing this one. @davemevans is correct.

When `EXT-X-MEDIA-SEQUENCE` tag is not present, clients should consider it is zero.

Even packagers like Shaka just generate this tag when media-sequence is not zero.

```
if (media_sequence_number > 0) {
        base::StringAppendF(&header, "#EXT-X-MEDIA-SEQUENCE:%d\n",
                            media_sequence_number);
      }
```

Thank you for your contribution @mojtaba-arvin

PR #261, opened 2021-10-01
Add ext-x-bitrate hls tag

DantasB:
- New tag included: '#EXT-X-BITRATE';
- New tests added.


m3u8/model.py
```diff
@@ -636,6 +647,9 @@ def dumps(self, last_segment):
         if self.byterange:
             output.append(',BYTERANGE=%s' % self.byterange)
 
+        if self.bitrate:
```
davemevans, line 650: Where is this attribute defined? I don't see it in the linked spec ...

DantasB: It's passed as parameter in the `__init__()` method and defined in the line 623. Is that your question?

davemevans: No, I meant where in https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-09 is the attribute BITRATE for a partial segment defined? Apologies for not being clearer.

DantasB: Oh, no problem :) It's inside the Media segment tags for the playlist tags https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.4.4.8

DantasB: So you're right. I think it should not be in the PartialSegment class.

PR #265, opened 2021-10-19
Add STABLE-VARIANT-ID attribute to StreamInfo

tassilucas:
Add EXT-X-STREAM-INF tag support to STABLE-VARIANT-ID attribute as specified in https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-09#section-4.4.6.2


leandromoreira:
there are also conflicts, can you solve them? cc @mauricioabreu

mauricioabreu:
Closed in favor of #286

PR #269, opened 2021-12-09
Add support for Content Steering tag

davemevans:
... and also PATHWAY-ID on STREAM-INF tags.

See https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis-10#section-4.4.6.6


leandromoreira (approved):

leandromoreira:
that's awesome thanks @davemevans sorry for the delayed response, @mauricioabreu can you chek this as well.

mauricioabreu (approved):

PR #282, opened 2022-07-01
Use millisecond accuracy for EXT-X-PROGRAM-DATE-TIME

bbayles:
In [RFC 8216, section 4.3.2.6](https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.6), we have this line about `EXT-X-PROGRAM-DATE-TIME`:
> It [SHOULD](https://datatracker.ietf.org/doc/html/rfc2119#section-3) indicate a time zone and fractional parts of seconds, to millisecond accuracy.

The two examples it gives are:
> YYYY-MM-DDThh:mm:ss.SSSZ
> 2010-02-19T14:54:23.031+08:00

Currently `m3u8` uses [`datetime.isoformat`](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat) with no arguments. This will produce timestamps with different accuracy depending on the input:
> `YYYY-MM-DDTHH:MM:SS.ffffff`, if `microsecond` is not 0
> `YYYY-MM-DDTHH:MM:SS`, if `microsecond` is 0

Alas, neither of these is "millisecond accuracy." In this PR, I change the `timespec` argument to `milliseconds`. This causes `m3u8` to emit timestamps with millisecond accuracy.

---

(Note: the current [draft](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.4.4.6) of the HLS spec says "*at least* millisecond accuracy, which would allow for microsecond accuracy)


mauricioabreu:
Awesome job @bbayles thank you for your hard work!

PR #286, opened 2022-07-15
Add support for stable variant/rendition IDs

vevv:
Adds support for `STABLE-VARIANT-ID` and `STABLE-RENDITION-ID` tags as per https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-11.

https://github.com/globocom/m3u8/pull/265 had merge conflicts and implementation issues (variant IDs not added to iframes, rendition IDs not added), requester seems inactive


mauricioabreu (approved):

mauricioabreu:
Amazing job @vevv

PR #288, opened 2022-07-25
Add support for EXT-X-CUE-OUT-CONT attributes in Segment.dumps

bbayles:
This PR allows `EXT-X-CUE-OUT-CONT` tags with "SCTE-35 Enhanced" ([ref](https://docs.aws.amazon.com/medialive/latest/ug/sample-manifests-hls.html#sample-manifests-hls-scte-35-enhanced)) to be parsed and dumped.

`m3u8.model.Segment` gets a new attribute, `scte35_elapsedtime`. This corresponds to the `ElapsedTime` parameter in the tag. The `Duration` and `SCTE35` attributes are already parsed and stored in `m3u8.model.Segment` objects.

The test suite already had examples of playlists with these attributes. Before, doing a `loads` and then a `dumps` would strip them out, leaving only `#EXT-X-CUE-OUT-CONT`. Now they will persist.


tests/playlists.py
```diff
@@ -684,11 +684,11 @@
 #EXTINF:10,
 #EXT-X-PROGRAM-DATE-TIME:2015-06-18T23:22:10Z
 1432451707508/ts/71737/sequence143474338.ts
-#EXT-X-CUE-OUT-CONT:CAID=0x000000002310E3A8,ElapsedTime=161,Duration=181
```
bbayles, line 687: I changed this playlist to make sure the plain `#EXT-X-CUE-OUT-CONT\n` output gets tested.

mauricioabreu:
Thank you for your contribution @bbayles 
I will have this PR reviewed by the end of the day

PR #290, opened 2022-08-01
Persist EXT-OATCLS-SCTE35 in Segment.dumps

### LINKED ISSUE #183: #EXT-OATCLS-SCTE35 tag is getting removed
According to documentation, this library supports the '#EXT-OATCLS-SCTE35' tag. But, 
when I used this parser to trim manifest files, SCTE tags were getting removed.

Below are the steps I follow,
1. Downloaded master manifest file[using python request]
2. Passed downloaded master manifest string to m3u8.loads method and got parsed master manifest object[say parsed_master_manifest]
3. Iterated over playlists of parsed_master_manifest
    a. Downloaded child manifest file[using python request]
    b. Passed downloaded child manifest string to m3u8.loads method and got parsed child manifest object[say parsed_child_manifest]
    c. Initialized Segmentlist[say desired_segments]
    d. Appended desired segments in  desired_segments
    e. Replaced original segments list with desired_segments
4. Store Master and all child manifest on local machine using m3u8.dump method
 
Please expedite this. and me know If I am doing anything wrong. I would much appreciate if any of you suggest a solution for this.

bbayles:
This PR enhances support for [SCTE-35 enhanced](https://docs.aws.amazon.com/medialive/latest/ug/sample-manifests-hls.html#sample-manifests-hls-scte-35-enhanced) manifests.

`Segment.oatcls_scte35` is now set for `Segment`s with a `EXT-OATCLS-SCTE35` tag. That property is also set for `Segment`s in the same `EXT-X-CUE-OUT` to `EXT-X-CUE-IN` block.

`Segment.dumps` now preserves `EXT-OATCLS-SCTE35` as well.

Fixes #183


m3u8/parser.py
```diff
@@ -111,10 +111,13 @@ def parse(content, strict=False, custom_tags_parser=None):
             state['cue_out'] = True
 
         elif line.startswith(protocol.ext_x_cue_out):
-            _parse_cueout(line, state, string_to_lines(content)[lineno - 2])
+            _parse_cueout(line, state)
```
bbayles, line 114: This change allowed me to simplify "cue out" parsing. We no longer require the "previous line" to be examined.

m3u8/protocol.py
```diff
@@ -25,7 +25,7 @@
 ext_x_cue_out_cont = '#EXT-X-CUE-OUT-CONT'
 ext_x_cue_in = '#EXT-X-CUE-IN'
 ext_x_cue_span = '#EXT-X-CUE-SPAN'
-ext_x_scte35 = '#EXT-OATCLS-SCTE35'
```
bbayles, line 28: `EXT-X-SCTE35` is actually a different tag. It's specified in [this document](https://www.scte.org/documents/5003/SCTE_35_2022.pdf).

m3u8/model.py
```diff
@@ -6,7 +6,13 @@
 import os
 import errno
 
-from m3u8.protocol import ext_x_start, ext_x_key, ext_x_session_key, ext_x_map
+from m3u8.protocol import (
```
mauricioabreu, line 9: Nice, thank you!

m3u8/parser.py
mauricioabreu: You are right! Much better.

m3u8/protocol.py
mauricioabreu: Indeed! Nice catch

mauricioabreu (approved):

PR #292, opened 2022-08-10
Use os.makedirs with exist_ok=True

### LINKED ISSUE #291: playlist.dump() fails when saving file to root folder on Windows.
I guess nobody saves m3u8 on root directory but r: is a ramdisk on my system and following code will fail

`playlist.dump("r:/aaa.m3u8")`

with access  denied error.

https://github.com/globocom/m3u8/blob/50963a865b09396302284a300ec8aff5e8317bbc/m3u8/model.py#L369-L376

so maybe add `exist_ok=True`? or an extra check for root folder?

bbayles:
Fixes #291

This PR updates `M3U8.dump` such that it uses `exist_ok=True` when creating directories with `os.makedirs`.

The intended behavior now obtains:
* if the directory already exists, it will be used.
* if not, it will be created.
* if there was some error in creating it, the associated exception will be raised.

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -368,12 +368,8 @@ class M3U8(object):
 
     def _create_sub_directories(self, filename):
         basename = os.path.dirname(filename)
-        try:
-            if basename:
-                os.makedirs(basename)
-        except OSError as error:
-            if error.errno != errno.EEXIST:
-                raise
+        if basename:
+            os.makedirs(basename, exist_ok=True)
 
 
 class Segment(BasePathMixin):
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -612,21 +612,18 @@ def test_dump_should_create_sub_directories(tmpdir):
 
     assert_file_content(filename, expected)
 
-def test_dump_should_raise_if_create_sub_directories_fails(tmpdir, monkeypatch):
-    def raiseOSError(*args):
-        raise OSError
+def test_dump_should_raise_if_create_sub_directories_fails(tmpdir):
+    # The first subdirectory is read-only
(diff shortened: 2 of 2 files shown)


m3u8/model.py
```diff
@@ -368,12 +368,8 @@ def dump(self, filename):
 
     def _create_sub_directories(self, filename):
         basename = os.path.dirname(filename)
-        try:
-            if basename:
-                os.makedirs(basename)
-        except OSError as error:
-            if error.errno != errno.EEXIST:
```
bbayles, line 375: [This block](https://github.com/python/cpython/blob/a69256527f93d2aa32e76658deab829e324d97b6/Lib/os.py#L226-L230) in the CPython source explains why this check isn't enough:

> Cannot rely on checking for EEXIST, since the operating system could give priority to other errors like EACCES or EROFS

In #291, `EACCES` is the error code rather than `EEXIST`.

mauricioabreu: Good to know! Thank you!

tests/test_model.py
```diff
@@ -612,21 +612,18 @@ def test_dump_should_create_sub_directories(tmpdir):
...
-    raised = False
-    try:
-        obj = m3u8.M3U8(playlists.SIMPLE_PLAYLIST)
-        obj.dump(str(tmpdir.join('subdir1', 'playlist.m3u8')))
-    except OSError as e:
-        raised = True
-    finally:
-        assert raised
-
+def test_dump_should_raise_if_create_sub_directories_fails(tmpdir):
```
mauricioabreu, line 615: Nice tests

mauricioabreu (approved):

PR #303, opened 2022-10-26
Updated ReadMe file

KuldeepBorkar:
Currently all the Supported tags mentioned in the `README.rst` file does not have links provided to them, therefore now I have added the links to them however I was not able to find the proper links to

* #EXT-X-CUE-OUT
* #EXT-X-CUE-OUT-CONT
* #EXT-X-CUE-IN
* #EXT-X-CUE-SPAN
* #EXT-OATCLS-SCTE35

that's why I have kept them as it is in the same format as mentioned above and for the tags I found the link I mentioned them as the rest of the tags with link as

*  "EXT-X-ALLOW-CACHE"_ (i.e. in place of this " this ` ticks)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,7 +57,7 @@ Supported tags
 * `#EXT-X-KEY`_
 * `#EXT-X-STREAM-INF`_
 * `#EXT-X-VERSION`_
-* #EXT-X-ALLOW-CACHE
+* `#EXT-X-ALLOW-CACHE`_
 * `#EXT-X-ENDLIST`_
 * `#EXTINF`_
 * `#EXT-X-I-FRAMES-ONLY`_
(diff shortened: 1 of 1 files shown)


mauricioabreu:
thank you @KuldeepBorkar

KuldeepBorkar:
@mauricioabreu I wanted to know, do we need to tackle some special issues for the hacktober-fest label? I came to know about hacktober-fest a little late and therefore checked once after this pull request was merged so it displayed that if the repository has participated in Hacktober-fest then on Pull Request there will be a Hacktober-fest label added so that it can be added as a valid Pull Request for the fest also🐱

KuldeepBorkar:
@mauricioabreu ??

PR #304, opened 2022-11-01
Comparisons to singletons - PEP8 style guide

EwertonBello:
* Code based on [PEP 8 Style Guide](https://www.python.org/dev/peps/pep-0008/)

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -817,7 +817,7 @@ class Playlist(BasePathMixin):
         self.base_uri = base_uri
 
         resolution = stream_info.get('resolution')
-        if resolution is None:
+        if resolution is not None:
             resolution = resolution.strip('"')
             values = resolution.split('x')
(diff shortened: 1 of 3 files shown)


leandromoreira (approved):

m3u8/model.py
```diff
@@ -817,7 +817,7 @@ def __init__(self, uri, stream_info, media, base_uri):
         self.base_uri = base_uri
 
         resolution = stream_info.get('resolution')
-        if resolution != None:
+        if resolution is None:
```
bbayles, line 820: Should this be `is not`?

leandromoreira: good catch

EwertonBello: > Should this be `is not`?

Thanks, it's fixed now!

leandromoreira (approved):

PR #305, opened 2022-11-01
refact: Remove useless else

EwertonBello:


leandromoreira (changes requested):
I think this changes were tackled at a previous PR.

m3u8/model.py
```diff
@@ -817,7 +817,7 @@ def __init__(self, uri, stream_info, media, base_uri):
         self.base_uri = base_uri
 
         resolution = stream_info.get('resolution')
-        if resolution != None:
+        if resolution is None:
```
rokam, line 820: Here you are changing the logic.

leandromoreira: good catch

EwertonBello:
> I think this changes were tackled at a previous PR.

Thanks, the commit was incorrect, it's fixed now!

m3u8/model.py
```diff
@@ -817,7 +817,7 @@ def __init__(self, uri, stream_info, media, base_uri):
         self.base_uri = base_uri
 
         resolution = stream_info.get('resolution')
-        if resolution is None:
+        if resolution != None:
```
leandromoreira, line 820: it should be `resolution is not None`? also isn't this related to your other open PR, I think you must split the code changes.

tests/test_model.py
```diff
@@ -207,8 +207,8 @@ def test_segment_envivio_scte35_attribute():
 
 def test_segment_unknown_scte35_attribute():
     obj = m3u8.M3U8(playlists.CUE_OUT_INVALID_PLAYLIST)
-    assert obj.segments[0].scte35 is None
-    assert obj.segments[0].scte35_duration is None
+    assert obj.segments[0].scte35 == None
+    assert obj.segments[0].scte35_duration == None
```
leandromoreira, line 211: same here, I think you're mixing your PRs

EwertonBello: I believe it is the first commit, it has already been fixed in the second commit

m3u8/parser.py
```diff
@@ -412,8 +412,7 @@ def _cueout_envivio(line, state):
     res = re.match('.*DURATION=(.*),.*,CUE="(.*)"', value)
     if res:
         return (res.group(2), res.group(1))
-    else:
-        return None
+    return None
```
davemevans, line 415: `None` will be returned implicitly, so this statement is only useful if it's necessary to be explicit for some reason

mauricioabreu: I think this case is one of those "explicit is better than implicit"

mauricioabreu (approved):

mauricioabreu:
@EwertonBello can you rebase your branch against master? It has some conflicts. After rebasing I'll merge it.

nedmax:
Opting for explicit

Issue #308, opened 2022-11-24
With seg.key but seg.key.iv is None


HFrost0:
When we need AES to decrypt segment, IV is important. In some m3u8 file, iv is not provided while can be generated by a specific rule which I found in [decrypt-m3u8-playlist-encrypted-with-aes-128-without-iv](https://stackoverflow.com/questions/50628791/decrypt-m3u8-playlist-encrypted-with-aes-128-without-iv)

Can we do this process in m3u8 lib instead of leave the seg.key.iv to None (I found this behavior in version 3.2.0)?

bbayles:
Here is a way to fall back to the Media Sequence if the Initialization Vector is missing:

```python
import m3u8
playlist_text = '...'  # Your text here
parsed_playlist = m3u8.loads(playlist_text)
seq_no = parsed_playlist.media_sequence or 0
for n, segment in enumerate(parsed_playlist.segments, seq_no):
    iv = segment.key.iv or f'0x{n:032x}'
    # do whatever you need to do with the IV
```

Reference: [RFC 8216, Section 5.2](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-5.2)

```
   An EXT-X-KEY tag with a KEYFORMAT of "identity" that does not have an
   IV attribute indicates that the Media Sequence Number is to be used
   as the IV when decrypting a Media Segment, by putting its big-endian
   binary representation into a 16-octet (128-bit) buffer and padding
   (on the left) with zeros.
```

leandromoreira:
as always thank you @bbayles

HFrost0:
Thanks for your solution, is there any plan to put this inside the `m3u8` lib instead of letting user do it by themself?

bbayles:
I opened https://github.com/globocom/m3u8/pull/310, which would make things a bit simpler.

mauricioabreu:
Thank you @bbayles

PR #310, opened 2022-12-16
Add Segment.media_Sequence

bbayles:
Re: [Issue 308](https://github.com/globocom/m3u8/issues/308), this PR adds a `media_sequence` attribute to `Segment` objects.

From RFC 8216:

```
Each segment in a Media Playlist has a unique integer Media Sequence
   Number.  The Media Sequence Number of the first segment in the Media
   Playlist is either 0 or declared in the Playlist ([Section 4.4.3.2](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.4.3.2)).
   The Media Sequence Number of every other segment is equal to the
   Media Sequence Number of the segment that precedes it plus one.
```

It's useful to have access to the Media Sequence number on the `Segment` object because this value can be used for the initialization vector (section 5.2). It's also useful for tracking changes between requests of live playlists.

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -174,6 +174,10 @@ class M3U8(object):
         for attr, param in self.simple_attributes:
             setattr(self, attr, self.data.get(param))
 
+        if self.media_sequence is not None:
+            for i, segment in enumerate(self.segments, self.media_sequence):
+                segment.media_sequence = i
+
         self.files = []
         for key in self.keys:
             # Avoid None key, it could be the first one, don't repeat them
@@ -448,7 +452,8 @@ class Segment(BasePathMixin):
                  cue_out_start=False, cue_in=False, discontinuity=False, key=None, scte35=None,
                  oatcls_scte35=None, scte35_duration=None, scte35_elapsedtime=None, asset_metadata=None,
                  keyobject=None, parts=None, init_section=None, dateranges=None, gap_tag=None,
-                 custom_parser_values=None):
+                 media_sequence=None, custom_parser_values=None):
+        self.media_sequence = media_sequence
         self.uri = uri
         self.duration = duration
         self.title = title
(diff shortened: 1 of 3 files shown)


m3u8/model.py
```diff
@@ -174,6 +174,9 @@ def _initialize_attributes(self):
         for attr, param in self.simple_attributes:
             setattr(self, attr, self.data.get(param))
 
+        for i, segment in enumerate(self.segments, self.media_sequence or 0):
```
leandromoreira, line 177: I learned something new here :) I didn't know `enumerate` would work with two sources :)

Just out of curiosity, why does it return first the number iterator and secondly the list iterator? (since they were passed in other order)

leandromoreira (approved):
Another amazing piece of work! ty man

m3u8/model.py
bbayles: Two tricks here.

First, enumerate has a two argument form: `iterable`, `start_number`. For us, `self.segments` is the iterable, since the `SegmentList` type can be iterated. We want to start counting from the media sequence number, so we use it as the second argument.

Second, the `self.media_sequence or 0` trick. I want to handle the case where `media_sequence` is `None` and treat that ias `0`, as per the HLS spec. The `or` operator takes the left argument if it's truthy (`None` isn't) and the right operator if it's not. 

```python
None or 0  # Picks 0
0 or 0  # Picks the second 0
1 or 0  # Picks 1
2 or 0  # Picks 2
# etc
```

leandromoreira: cool ty

mauricioabreu:
Thank you @bbayles amazing work!
Do you mind rebasing your PR? I'm ready to merge it.

bbayles:
OK, this is rebased. Thanks!

mauricioabreu:
Thank you, @bbayles merged.

PR #313, opened 2023-03-09
Rework segment.key as segment.keys, a list of keys

### LINKED ISSUE #283: Parsing and dumping of multiple EXT-X-KEYs per segment is not supported
The HLS spec is clear that a segment may be preceeded with multiple EXT-X-KEY tags, so long as they resolve to a single decryption key. This allows multiple DRMs to be supported using Common Encryption.

When the example playlist below is parsed, `segment[0]['key']` only lists the last key that preceeded the segment. Dumping the playlist object results in only the last key preceeding the segment appearing in the output. We need to be able to have a list of keys associated with a segments, possibly manipulate these, and dump them correctly.

Note: `m3u8['keys']` includes all keys, but it doesn't correctly associated keys with segments using `by_key`, presumably because of the above.

Usually I would knock something together, but key handling seems a bit confusing to me and am looking for a bit of direction about what might be the best way to approach this. I am tempted simply to add `segment['keys']` and always return a list (or None when no keys), but it's not obvious what `segment['key']` should be, or how `m3u8['keys']` would operate, in this model. Does anyone else have an opinion here?

```
#EXTM3U
#EXT-X-VERSION:6
#EXT-X-TARGETDURATION:6
#EXT-X-MAP:URI="0.m4s"
#EXT-X-PROGRAM-DATE-TIME:2022-07-04T11:30:11.680150Z
#EXT-X-KEY:METHOD=SAMPLE-AES,URI="skd://blah",IV=0xd7c3306f79c9b7d179f5dd09e2431d6e,KEYFORMAT="com.apple.streamingkeydelivery",KEYFORMATVERSIONS="1"
#EXT-X-KEY:METHOD=SAMPLE-AES,URI="data:pssh...",IV=0xd7c3306f79c9b7d179f5dd09e2431d6e,KEYFORMAT="urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed",KEYFORMATVERSIONS="1"
#EXT-X-KEY:METHOD=SAMPLE-AES,URI="data:pssh...",IV=0xd7c3306f79c9b7d179f5dd09e2431d6e,KEYFORMAT="com.microsoft.playready",KEYFORMATVERSIONS="1"
#EXTINF:6.006,
275879823.m4s
```

rlaphoenix:
This closes #283 as it now follows RFC and keeps all preceding keys for the segment.

- `m3u8_obj.keys` still has the same behaviour.
- `m3u8_obj.segments[0].key` is now removed.
- `m3u8_obj.segments[0].keys` is added in it's place, which will now always be a list.
- The list will always be present, it will never be `None` anymore, and if the segment explicitly has no DRM by specifying an EXT-X-KEY of METHOD=NONE then it will be a list with that key.
- It does not check if there's a METHOD=NONE with other conflicting EXT-X-KEY information.
- It still follows the same behaviour of not duplicating the EXT-X-KEY information on future segments when dumping.

Do note that it only clears the list of Keys when it reaches an EXT-X-DISCONTINUITY, or it has reached a ts segment.

The copy() statement is used because it clears after the segment is appended to `data`, yet the .clear() to `current_keys` follows in the appended `segment` unless we copy that data in memory prior.

I don't think there's a need for `.by_key()` or `m3u8_obj.keys` anymore, but I've made sure those still work as intended in case there's a different reason those exist.

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -167,7 +167,12 @@ class M3U8(object):
         self.segment_map = [InitializationSection(base_uri=self.base_uri, **params) if params else None
                      for params in self.data.get('segment_map', [])]
         self.segments = SegmentList([
-            Segment(base_uri=self.base_uri, keyobject=find_key(segment.get('key', {}), self.keys), **segment)
+            Segment(
+                base_uri=self.base_uri,
+                keyobjects=[
+                    find_key(segment_key, self.keys)
+                    for segment_key in segment['keys']],
+                **segment)
             for segment in self.data.get('segments', [])
         ])
 
@@ -430,8 +435,8 @@ class Segment(BasePathMixin):
     `byterange`
       byterange attribute from EXT-X-BYTERANGE parameter
 
-    `key`
-      Key used to encrypt the segment (EXT-X-KEY)
+    `keys`
+      Keys used to encrypt the segment (EXT-X-KEY)
 
     `parts`
       partial segments that make up this segment
@@ -448,9 +453,9 @@ class Segment(BasePathMixin):
 
     def __init__(self, uri=None, base_uri=None, program_date_time=None, current_program_date_time=None,
                  duration=None, title=None, bitrate=None, byterange=None, cue_out=False,
-                 cue_out_start=False, cue_in=False, discontinuity=False, key=None, scte35=None,
+                 cue_out_start=False, cue_in=False, discontinuity=False, keys=None, scte35=None,
                  oatcls_scte35=None, scte35_duration=None, scte35_elapsedtime=None, asset_metadata=None,
-                 keyobject=None, parts=None, init_section=None, dateranges=None, gap_tag=None,
+                 keyobjects=None, parts=None, init_section=None, dateranges=None, gap_tag=None,
                  media_sequence=None, custom_parser_values=None):
         self.media_sequence = media_sequence
         self.uri = uri
@@ -470,7 +475,7 @@ class Segment(BasePathMixin):
         self.scte35_duration = scte35_duration
         self.scte35_elapsedtime = scte35_elapsedtime
         self.asset_metadata = asset_metadata
-        self.key = keyobject
+        self.keys = keyobjects
         self.parts = PartialSegmentList( [ PartialSegment(base_uri=self._base_uri, **partial) for partial in parts ] if parts else [] )
         if init_section is not None:
             self.init_section = InitializationSection(self._base_uri, **init_section)
@@ -486,14 +491,9 @@ class Segment(BasePathMixin):
     def dumps(self, last_segment, timespec='milliseconds'):
         output = []
 
-        if last_segment and self.key != last_segment.key:
-            output.append(str(self.key))
-            output.append('\n')
-        else:
-            # The key must be checked anyway now for the first segment
-            if self.key and last_segment is None:
-                output.append(str(self.key))
-                output.append('\n')
+        if not last_segment or (self.keys and self.keys != last_segment.keys):
+            for key in self.keys:
+                output.append(str(key) + '\n')
 
         if last_segment and self.init_section != last_segment.init_section:
             if not self.init_section:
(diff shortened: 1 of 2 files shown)


mauricioabreu:
Hey @rlaphoenix sorry for the late reply. I will review it this weekend

rlaphoenix:
Ok great. One thing I have thought of is we could keep compatibility by re-adding back `key`, and have it be the last entry of `keys`, sort of like how it works currently in master. That way current code expecting `key` that understood its flaw will still work and can gracefully migrate to using `keys`. Maybe mark it as deprecated and remove it 6 months down the line.

mauricioabreu (approved):

mauricioabreu:
@rlaphoenix good job!

About implementing backwards compatibility, I think we can upgrade the major version of `m3u8`, because a major is expected to break clients that don't follow the new changes.

mauricioabreu:
@rlaphoenix it looks like the test suite failed. Can you check it?

rlaphoenix:
Hi, apologize for all the squashes, but I believe it have the test's fixed. However, the tests on Windows causes it to compare e..g, `/foo/bar` with `\foo\bar` or `\\foo\\bar`, e.t.c. Therefore I currently have 11 failed tests, most being caused by that. However, I also have a few tests related to by_key() failing.

```
============================================================================== short test summary info ===============================================================================
FAILED tests/test_loader.py::test_load_should_create_object_from_file_with_relative_segments - AssertionError: assert 'C:\\Users\\Justin\\Projects\\m3u8\\tests/key.bin' == 'C:\\Users\\Justin\\Projects\\m3u8\\tests\\key.bin'
FAILED tests/test_loader.py::test_load_should_create_object_from_uri_with_relative_segments - AssertionError: assert 'http://localhost:8112\\path/key.bin' == 'http://localhost:8112/path/key.bin'
FAILED tests/test_model.py::test_keys_on_simple_encrypted_playlist - assert 2 == 1
FAILED tests/test_model.py::test_segment_attribute_with_multiple_keys - AttributeError: 'NoneType' object has no attribute 'uri'
FAILED tests/test_model.py::test_dump_should_raise_if_create_sub_directories_fails - Failed: DID NOT RAISE <class 'OSError'>
FAILED tests/test_model.py::test_length_segments_by_key - assert 0 == 2
FAILED tests/test_model.py::test_list_segments_by_key - AssertionError: assert '' == '../../../../hls/streamNum82400.ts\n../../../../hls/streamNum82401.ts'
FAILED tests/test_model.py::test_replace_segment_key - assert '#EXTM3U\n#EXT-X-MEDIA-SEQUENCE:82400\n#EXT-X-ALLOW-CACHE:NO\n#EXT-X-VERSION:2\n#EXT-X-TARGETDURATION:8\n#EXTINF:8,\n../../../../hls/streamNum82400.ts\n#EXTINF:8,\n../../....
FAILED tests/test_model.py::test_m3u8_should_propagate_base_uri_to_segments - AssertionError: assert '/any/path/entire1.ts' == '\\any\\path\\entire1.ts'
FAILED tests/test_model.py::test_m3u8_should_propagate_base_uri_to_key - AssertionError: assert '/any/key.bin' == '\\any\\key.bin'
FAILED tests/test_model.py::test_m3u8_should_propagate_base_uri_to_session_key - AssertionError: assert '/any/key.bin' == '\\any\\key.bin'
========================================================= 11 failed, 218 passed, 1 skipped, 9 xfailed, 2 warnings in 11.91s ==========================================================
```

mauricioabreu:
@rlaphoenix If I understood it right, your code is backwards compatible right? Maybe in a major update release we can remove the `key` and stick with the `keys` only

rlaphoenix:
> @rlaphoenix If I understood it right, your code is backwards compatible right? Maybe in a major update release we can remove the `key` and stick with the `keys` only

Correct. Pulling this as it is right now might not be the best idea. As stated before there's no real tests made for `keys` yet with some tests testing `key` still, and some modified to test `keys`. Some tests are failing like the `by_key` and stuff.

It's sort of up to you what you want to do. We could just drop `key` and the `by_key()` function, update the tests to work with `keys`. Or we could write new tests for `keys` and fix the remaining broken tests (e.g., by_key, which I feel should still go).

One thing I will say is that I use these current changes (before the squashes/changes to add back `key`) and multiple people use it. It has been working like a charm with no problems requiring changes to this pull, and minimal changes to the projects that used `key`.

PR #314, opened 2023-03-22
Add support for ext x image tags

### LINKED ISSUE #287: Add support to tag #EXT-X-IMAGE
Image Media Playlist for HLS (also known as Thumbnails) introduced a new tag #EXT-X-IMAGE in HLS Playlist.
Roku is one example of a platform that implements this specification which is located here:
[https://github.com/image-media-playlist/spec/blob/master/image_media_playlist_v0_4.pdf](https://github.com/image-media-playlist/spec/blob/master/image_media_playlist_v0_4.pdf)
This tag #EXT-X-IMAGE is not supported in Python m3u8 Parser.

darkbaboon:
Add support for Image Playlist with #EXT-X-IMAGE-STREAM-INF, #EXT-IMAGE-ONLY and #EXT-X-TILES tags
Resolves #287


mauricioabreu:
thank you @darkbaboon but there are still some pending conversations. Also, I pushed a commit minutes ago (conflicts again, sorry)

PR #315, opened 2023-05-02
Avoid using decimal.Decimal for integer values

bbayles:
I did some profiling of this module with `cProfile` and found that the most expensive part of the `dumps` function is the `number_to_string`, which uses the `decimal` module.

This PR changes removes few calls to `number_to_string` where it's possible:
* `EXT-X-DISCONTINUITY-SEQUENCE` is necessarily an integer, per RFC 8216
* `LAST-MSN`
* `SKIPPED-SEGMENTS`


mauricioabreu (approved):
Awesome, thank you!

PR #263, opened 2021-10-05
Fix URLs join failed when path contains `//`

SeaHOH:
close #245

PR #316, opened 2023-05-03
Re-work URL joining to follow RFC 3986

bbayles:
This PR updates how the `absolute_uri` method for objects that use the `BasePathMixin`.

I think the changes related to issue #245 and PR #263 introduced some unusual and surprising behavior. The issue related to handling relative URLs that start with `//`, like `//segment.ts`.

Joining URLs is covered by [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986.html), which specifically describes the `//` case. From [section 5.4.1](https://www.rfc-editor.org/rfc/rfc3986.html#section-5.4.1):
image

Python's built-in `urllib.parse.urljoin` produces the expected result, but `m3u8.parse.urljoin` does not:
```python
>>> from urllib.parse import urljoin as python_urljoin
>>> from m3u8.parser import urljoin as m3u8_urljoin

>>> m3u8_urljoin('http://a/b/c/d;p?q', '//g')
'http://a//g'

>>> python_urljoin('http://a/b/c/d;p?q', '//g')
'http://g'
```

This PR updates `m3u8` to prefer the standards-based approach.

### DIFF
--- a/m3u8/parser.py
+++ b/m3u8/parser.py
@@ -594,20 +594,6 @@ def normalize_attribute(attribute):
     return attribute.replace('-', '_').lower().strip()
 
 
-def is_url(uri):
-    return uri.startswith(URI_PREFIXES)
-
-
-def urljoin(base, url):
-    base = base.replace('://', '\1')
-    url = url.replace('://', '\1')
-    while '//' in base:
-        base = base.replace('//', '/\0/')
-    while '//' in url:
-        url = url.replace('//', '/\0/')
-    return _urljoin(base.replace('\1', '://'), url.replace('\1', '://')).replace('\0', '')
-
-
 def get_segment_custom_value(state, key, default=None):
     """
     Helper function for getting custom values for Segment
--- a/m3u8/parser.py
+++ b/m3u8/parser.py
@@ -15,7 +15,6 @@ http://tools.ietf.org/html/draft-pantos-http-live-streaming-08#section-3.2
 http://stackoverflow.com/questions/2785755/how-to-split-but-ignore-separators-in-quoted-strings-in-python
 '''
 ATTRIBUTELISTPATTERN = re.compile(r'''((?:[^,"']|"[^"]*"|'[^']*')+)''')
-URI_PREFIXES = ('https://', 'http://', 's3://', 's3a://', 's3n://')
 
 def cast_date_time(value):
     return iso8601.parse_date(value)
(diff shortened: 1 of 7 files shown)


m3u8/parser.py
```diff
@@ -15,7 +15,6 @@
 http://stackoverflow.com/questions/2785755/how-to-split-but-ignore-separators-in-quoted-strings-in-python
 '''
 ATTRIBUTELISTPATTERN = re.compile(r'''((?:[^,"']|"[^"]*"|'[^']*')+)''')
-URI_PREFIXES = ('https://', 'http://', 's3://', 's3a://', 's3n://')
```
bbayles, line 18: This change allows playlists to use additional URL schemes without having to update this list. [RFC 8216](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.4.6.6) does mention at least the `data://` scheme.

tests/test_loader.py
```diff
@@ -42,7 +42,7 @@ def test_load_should_create_object_from_file_with_relative_segments():
     obj = m3u8.load(playlists.RELATIVE_PLAYLIST_FILENAME)
     expected_key_abspath = '%s/key.bin' % os.path.dirname(base_uri)
     expected_key_path = '../key.bin'
-    expected_ts1_abspath = '%s/entire1.ts' % base_uri
+    expected_ts1_abspath = '/entire1.ts'
```
bbayles, line 45: I think files with a leading `/` were being mis-handled as well.

On a file system, joining `/tmp/path/to/somewhere` with `/entire.ts` should produce `/entire.ts`.

```python
>>> from os.path import join as python_join
>>> python_join('/path/to/somewhere', '/entire1.ts')
'/entire1.ts'
```

mauricioabreu: Agreed

m3u8/parser.py
mauricioabreu: Awesome! You remove a line while adding a new feature 👏🏼

SeaHOH:
Greatly surprised. So, `//` should not be in a path string? But I have saw someone did this.

mauricioabreu:
@SeaHOH maybe Windows paths?

bbayles:
I think it's fine to use `//`, it just needs to mean something standard.

SeaHOH:
A HTTP URL, just like this: `https://some.domain//path/to/segmentx.ts`, the `//` is in head of the relative path entires, not a host name.

bbayles:
I think that particular absolute URL is fine.

The question is, what do you do when you need to have the starting URL `http://a/b/c/d;p?q` and then are given the relative URL  `//g`?

The answer, going back to 1995 with RFC 1808, has been consistently the same:
* [RFC 1808](https://www.rfc-editor.org/rfc/rfc1808.txt)
* [RFC 2396](https://www.rfc-editor.org/rfc/rfc2396.txt)
* [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986.txt)

Issue #326, opened 2023-09-08
Wrongfully parsing custom tag as URI instead of ignoring it


MrChick:
The readme states the following:

> Quoting the documentation:
> 
> > Lines that start with the character '#' are either comments or tags.
> > Tags begin with #EXT.  They are case-sensitive.  All other lines that
> > begin with '#' are comments and SHOULD be ignored.
> 
> This library ignores all the non-standard tags by default.

I have a m3u file from my cable router (DVBC streaming) using the following format:

> #EXTM3U
> #EXTINF:0,Channel 1
> #EXTVLCOPT:network-caching=1000
> rtsp://192.168.[...]
> #EXTINF:0,Channel 2
> #EXTVLCOPT:network-caching=1000
> rtsp://192.168.[...]
> #EXTINF:0,Channel 3
> #EXTVLCOPT:network-caching=1000
> rtsp://192.168.[...]

It looks like [parser.py](https://github.com/globocom/m3u8/blob/ed0afc08981a0239543860a8790d8a49581fa083/m3u8/parser.py#L79C3-L79C3) expects any line following an EXTINF that is not a handled tag to be the actual URI no matter what.

I was expecting the EXTVLCOPT tags to simply be ignored, but instead after parsing the m3u file I get the channels correctly as titles, but instead of the rtsp URIs i get the EXTVLCOPT lines returned as URI.

I managed to easily remedy this by using a custom tag parser, but it sounds to me like the EXTVLCOPT tag should have been ignored out of the box?

bbayles:
I think PR #327 will fix this.

```
In [1]: from m3u8 import loads

In [2]: playlist_text = """#EXTM3U
   ...: #EXTINF:0,Channel 1
   ...: #EXTVLCOPT:network-caching=1000
   ...: rtsp://192.168.[...]
   ...: #EXTINF:0,Channel 2
   ...: #EXTVLCOPT:network-caching=1000
   ...: rtsp://192.168.[...]
   ...: #EXTINF:0,Channel 3
   ...: #EXTVLCOPT:network-caching=1000
   ...: rtsp://192.168.[...]"""

In [3]:

In [3]: parsed_playlist = loads(playlist_text, uri="https://localhost:24601/hls/playlist.m3u8")

In [4]: parsed_playlist.data
Out[4]:
{'media_sequence': 0,
 'is_variant': False,
 'is_endlist': False,
 'is_i_frames_only': False,
 'is_independent_segments': False,
 'is_images_only': False,
[...]

PR #327, opened 2023-09-08
Issue 326: allow tags after EXTINF

bbayles:
Re: issue #326, this PR updates the parser code such that a URL is not required after the `EXTINF` tag.


mauricioabreu:
How nice! Thank you for the quick and awesome fix @bbayles

PR #333, opened 2023-10-13
Use built-in ISO 8601 parsing

### LINKED ISSUE #331: Use built-in ISO 8601 parsing
Continuing in the vein of #38, the [built-in parsing of ISO 8601 that was added in Python 3.7](https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat) should be used. This would require bumping the minimum supported Python version, but 3.8 is the oldest version that's still supported, so that should be a fine thing to do. I'd be glad to submit a PR for this, if it will be accepted

daveisfera:
Fixes #331


daveisfera:
[`datetime.fromisoformat()` had limited parsing until 3.11](https://bugs.python.org/issue35829), so looks like this needs to wait

leandromoreira:
that's a bummer :/ nice finding @daveisfera thanks

PR #340, opened 2023-10-31
Support EXT-X-CUE-OUT:Elapsed/Duration format

bbayles:
This PR updates the `EXT-X-CUE-OUT` parsing function to support a format that I encountered in the wild.

Instead of the "SCTE-35 enhanced" style, which is already supported:
```
#EXT-X-CUE-OUT-CONT:ElapsedTime=$ELAPSEDTIME,Duration=$DURATION,SCTE35=$SCTE35
```

This has:
```
#EXT-X-CUE-OUT-CONT:$ELAPSEDTIME/$DURATION
```


mauricioabreu:
@bbayles sorry for the long delay. I'll do the best I can so it does happen again.

Good job!

Issue #220, opened 2020-07-10
detect and raise version if specified too low


thenewguy:
If the version is specified in a playlist that is too low, it should be raised when converted back to a string.

```
The following features aren't backward compatible. Older clients may fail to play the content if you use these features but don't specify the protocol version where they were introduced:

You must use at least protocol version 2 if you have IV in EXT-X-KEY.

You must use at least protocol version 3 if you have floating point EXTINF duration values.

You must use at least protocol version 4 if you have EXT-X-BYTERANGE or EXT-X-IFRAME-ONLY.

You must use at least protocol version 5 if you specify the SAMPLE-AES encryption method in EXT-X-KEY, or if you have KEYFORMAT and KEYFORMATVERSIONS attributes in EXT-X-KEY, or if you have EXT-X-MAP.

You must use at least protocol version 6 if you have EXT-X-MAP in a Media Playlist that does not contain EXT-X-I-FRAMES-ONLY.

You must use at least protocol version 7 if you specify "SERVICE" values for the INSTREAM-ID attribute of EXT-X-MEDIA.

You must use at least protocol version 8 if you use variable substitution.
```

Ref:
- https://developer.apple.com/documentation/http_live_streaming/about_the_ext-x-version_tag

thenewguy:
I see there is an [expected failure for this](https://github.com/globocom/m3u8/blob/b99fce0739f57d9b020c44f5ebae1727eee5f225/tests/test_strict_validations.py#L29) - is that because it will not be fixed or that it is waiting implementation?

leandromoreira:
I don't know @thenewguy it seems very reasonable that m3u8 would check the version matching featuress... but I'm afraid it can broken current usages?!... maybe we could do that but let it off and only turn on when strict is True, what do you think?

cc @mauricioabreu @flavioribeiro @igorsobreira 

PS: also maybe this `VersionMatching` thing should be placed into a self component (like a class...) instead of within the currrent parser, it can be used by the parser but not developed there.

mauricioabreu:
I think it is ok to make it work when `strict` is on.

mauricioabreu:
Closed via #366

PR #341, opened 2023-11-01
feat: add version_matching module

davigsousa:
As described at #220 

I implemented an initial version of a version_matching module.
- As suggested by @leandromoreira, it is a separated module from parser.
- I added some version matching rules from the reference specified by @thenewguy. I hope we can add the following ones in next PRs.
- It only runs in strict mode


m3u8/version_matching.py
```diff
@@ -0,0 +1,43 @@
...
+def valid_in_all_rules(line_number: int, line: str, version: float):
+    rules: List[Callable[[str, float], bool]] = [
+        valid_iv_in_EXT_X_KEY,
+        valid_floating_point_EXTINF,
+        valid_EXT_X_BYTERANGE_or_EXT_X_I_FRAMES_ONLY,
+    ]
+
+    for rule in rules:
+        if not rule(line, version):
+            raise VersionMatchingError(line_number, line)
```
leandromoreira, line 34: I like the whole idea. Thanks for the work! I'd suggest adding why it failed so developers can know how to fix it.

Something in the lines of:

```python
    rules: List[Callable[[str, float], bool]] = [
        valid_iv_in_EXT_X_KEY,
        valid_floating_point_EXTINF,
        valid_EXT_X_BYTERANGE_or_EXT_X_I_FRAMES_ONLY,
    ]

    for rule in rules:
        if not rule(line, version):
            raise VersionMatchingError(line_number, line, f"failed to run the checker {rule.__name__}")
 # or even using a more descriptive help; encoding rules as a dict, k -> v, k being the string describing the check and the value being the function/callback
# maybe add the fields reason and how_to_fix attached to each validation, I dunno
```

Also, maybe we should have a list of errors wrapped into a final exception so the user can know a list of what to fix instead of running it multiple times.

```python
errors = []
# ...
if len(errors) > 0:
  raise ManifestInvalid(... , errors)
```

m3u8/version_matching_rules.py
```diff
@@ -0,0 +1,44 @@
...
+    if not protocol.ext_x_key in line:
+        return True
+
+    if "IV" in line:
+        return version >= 2
+
+    return True
+
+
+# You must use at least protocol version 3 if you have floating point EXTINF duration values.
```
leandromoreira, line 15: maybe use this tip/description attached to the error helping users to know how to fix their problems

m3u8/parser.py
```diff
@@ -75,8 +75,12 @@ def parse(content, strict=False, custom_tags_parser=None):
         "current_segment_map": None,
     }
 
+    lines = string_to_lines(content)
+    if strict:
```
leandromoreira, line 79: where does the symbol `strict` come from? is it from args? env?

m3u8/version_matching.py
davigsousa: Thank you! I agree with you, I'll work on that!

davigsousa: Done! Could you please take a look on that?

m3u8/version_matching_rules.py
davigsousa: Got it!

davigsousa:
All erros are now raised when we finish to validate a file in strict mode, example:
```py
Traceback (most recent call last):
  File "/home/davi/projects/m3u8/./invalid.py", line 4, in <module>
    content = m3u8.parse(invalid_versioned_playlists.M3U8_RULE_IV, strict=True)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/davi/projects/m3u8/m3u8/parser.py", line 82, in parse
    raise Exception(found_errors)
Exception: [VersionMatchingError(line_number=2, line='#EXT-X-KEY: METHOD=AES-128, IV=0x123456789ABCDEF0123456789ABCDEF0, URI="https://example.com/key.bin"', how_to_fix='Change the protocol version to 2 or higher.', description='You must use at least protocol version 2 if you have IV in EXT-X-KEY.'), VersionMatchingError(line_number=4, line='#EXTINF: 10.0,', how_to_fix='Change the protocol version to 3 or higher.', description='You must use at least protocol version 3 if you have floating point EXTINF duration values.')]
```

m3u8/version_matching.py
```diff
@@ -0,0 +1,39 @@
...
+        if not validator.validate():
+            errors.append(validator.get_error())
+
+    return errors
+
+
+def validate(file_lines: List[str]) -> List[VersionMatchingError]:
+    found_version = get_version(file_lines)
+    if found_version is None:
+        return []
```
leandromoreira, line 32: For master variant is it mandatory to have version? if that's the case then I think we should return an error here stating that the master playlist doesn't contain a version.

Again we can leave that for a future PR if you wish.

m3u8/version_matching_rules.py
```diff
@@ -0,0 +1,109 @@
...
+from typing import List
+
+from m3u8 import protocol
+
+
+@dataclass
+class VersionMatchingError(Exception):
+    line_number: int
+    line: str
+    how_to_fix: str = "Please fix the version matching error."
```
leandromoreira, line 11: ❤️

m3u8/version_matching_rules.py
```diff
@@ -0,0 +1,109 @@
...
+            return True
+
+        chunks = self.line.replace(protocol.extinf + ":", "").split(",", 1)
+        duration = chunks[0]
+
+        def is_number(value: str):
+            try:
+                float(value)
+                return True
+            except:
```
leandromoreira, line 79: nit: we should catch `ValueError`

leandromoreira (approved):
Outstanding work you've done here, thanks a lot. I left some comments/nit but if you want, you can handle them in a new PR.

leandromoreira:
> All errors are now raised when we finish to validate a file in strict mode, example:
> 
> ```python
> Traceback (most recent call last):
>   File "/home/..././invalid.py", line 4, in <module>
>     content = m3u8.parse(invalid_versioned_playlists.M3U8_RULE_IV, strict=True)
>               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/.../m3u8/m3u8/parser.py", line 82, in parse
>     raise Exception(found_errors)
> Exception: [VersionMatchingError(line_number=2, line='#EXT-X-KEY: METHOD=AES-128, IV=0x123456789ABCDEF0123456789ABCDEF0, URI="https://example.com/key.bin"', how_to_fix='Change the protocol version to 2 or higher.', description='You must use at least protocol version 2 if you have IV in EXT-X-KEY.'), VersionMatchingError(line_number=4, line='#EXTINF: 10.0,', how_to_fix='Change the protocol version to 3 or higher.', description='You must use at least protocol version 3 if you have floating point EXTINF duration values.')]
> ```
> 
> I thought about printing them, but I didn't see any other prints in Parse, so I just raised

I think raising it's the best we can do, I suggested we use the group exception so the natural print will be easier to read.

m3u8/parser.py
```diff
@@ -74,8 +74,15 @@ def parse(content, strict=False, custom_tags_parser=None):
...
+        found_errors = version_matching.validate(lines)
+
+        if len(found_errors) > 0:
+            raise Exception(found_errors)
```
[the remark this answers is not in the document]
davigsousa: I wanted to use [ExceptionGroup](https://docs.python.org/3/library/exceptions.html#ExceptionGroup), but it is from Python 3.11 and the project requires only Python3.7+, I dont know if we can use it here.

davigsousa:
> I suggested we use the group exception

As I said, I wanted to use [ExceptionGroup](https://docs.python.org/3/library/exceptions.html#ExceptionGroup), but it is from Python 3.11 and the project requires only Python3.7+, I dont know if we can use it here. 😕

leandromoreira:
> > I suggested we use the group exception
> 
> As I said, I wanted to use [ExceptionGroup](https://docs.python.org/3/library/exceptions.html#ExceptionGroup), but it is from Python 3.11 and the project requires only Python3.7+, I dont know if we can use it here. 😕

that's okay, thanks for the efforts

leandromoreira:
A minor feature we can add, if the verification list keep growing, is a map like structure to avoid N*N:

```python

if hash(line) in verifications:
  verifications[verification_hash(line)].valid?

def verification_hash(line):
  if line.contains("version"):
    return "version"
```

mauricioabreu:
Awesome work, @davigsousa

leandromoreira:
It looks like related to:

```python
available_rules: List[type[VersionMatchRuleBase]] = []
```

Maybe replacing `type[` by `typing.Type` or `type(`.

https://docs.python.org/3.7/library/typing.html?highlight=type#typing.Type

```bash
tests/test_invalid_versioned_playlists.py:9: in <module>
    import m3u8
m3u8/__init__.py:9: in <module>
    from m3u8.model import (
m3u8/model.py:8: in <module>
    from m3u8.parser import format_date_time, parse
m3u8/parser.py:17: in <module>
    from m3u8 import protocol, version_matching
m3u8/version_matching.py:4: in <module>
    from m3u8.version_matching_rules import VersionMatchingError, available_rules
m3u8/version_matching_rules.py:108: in <module>
    ValidEXTXBYTERANGEOrEXTXIFRAMESONLY,
E   TypeError: 'type' object is not subscriptable
```

tests/test_invalid_versioned_playlists.py
```diff
@@ -0,0 +1,24 @@
...
+
+import invalid_versioned_playlists
+import pytest
+
+import m3u8
+
+
+@pytest.mark.xfail
+def test_should_fail_if_iv_in_EXT_X_KEY_and_version_less_than_2():
+    m3u8.parse(invalid_versioned_playlists.M3U8_RULE_IV)
```
leandromoreira, line 14: nit: maybe we should assert the exception or at least partially the error message/help

```python
def test_raises():
    with pytest.raises(Exception) as exc_info:   
        raise Exception('some info')
    # these asserts are identical; you can use either one   
    assert exc_info.value.args[0] == 'some info'
    assert str(exc_info.value) == 'some info'
```

ref https://stackoverflow.com/a/23514853

leandromoreira: ref https://docs.pytest.org/en/7.1.x/how-to/assert.html

leandromoreira: ref https://pytest-with-eric.com/introduction/pytest-assert-exception/

leandromoreira (commented):
good work

mauricioabreu:
Hello @davigsousa, are you still working on these awesome changes? If you don't have time to complete the task, I can take over the PR from here and handle the finalization and merging.

davigsousa:
Hi @mauricioabreu , unfortunately I have very little time for open source lately, sorry for that! I wanted to finish the work here, but thank you for your help, I'll accept it! We could speed up the merge here.

You can take over, thank you very much!

PR #343, opened 2023-11-09
[POC] Add default duration

radujica:


mauricioabreu:
Hello @radujica 

`Duration` of a data segment should be mandatory. I'm not a big fan of adding default values. What is your opinion @leandromoreira ?

mauricioabreu:
I think the best option here is adding a better message. Since this library is a parser, we should not be adding default values. I hope you understand it.

PR #356, opened 2024-02-06
Parse as float and then round to int

### LINKED ISSUE #355: RESOLUTION as float raises exception
The `RESOLUTION` value in `EXT-X-STREAM-INF` is listed as being a decimal value so some outputs list it with `.0` and that causes the parsing to fail ( see https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.4.2 )

daveisfera:
Fixes #355

### DIFF
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -959,7 +959,7 @@ class Playlist(BasePathMixin):
         if resolution is not None:
             resolution = resolution.strip('"')
             values = resolution.split("x")
-            resolution_pair = (int(values[0]), int(values[1]))
+            resolution_pair = (int(round(float(values[0]))), int(round(float(values[1]))))
         else:
             resolution_pair = None
 
--- a/m3u8/model.py
+++ b/m3u8/model.py
@@ -959,7 +959,10 @@ class Playlist(BasePathMixin):
         if resolution is not None:
             resolution = resolution.strip('"')
             values = resolution.split("x")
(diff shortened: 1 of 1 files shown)


m3u8/model.py
```diff
@@ -959,7 +959,10 @@ def __init__(self, uri, stream_info, media, base_uri):
         if resolution is not None:
             resolution = resolution.strip('"')
             values = resolution.split("x")
-            resolution_pair = (int(values[0]), int(values[1]))
+            resolution_pair = (
+                int(round(float(values[0]))),
```
leandromoreira, line 963: just out of curiosity: is there any packager outputting manifests with decimal resolution?

leandromoreira (approved):

m3u8/model.py
daveisfera: I had some example files I was testing with that had `1280.0x720.0` but I'm not sure where they came from. `ffmpeg` [outputs them as integers](https://github.com/FFmpeg/FFmpeg/blob/e38092ef9395d7049f871ef4d5411eb410e283e0/libavformat/hlsplaylist.c#L86) and I assume that most do as well

leandromoreira: no worries, thank you

leandromoreira:
H264 uses integers to represent resolution. `pic_width_in_mbs_minus_1 ue(v)`
image

daveisfera:
Closing since this is out of spec

PR #359, opened 2024-03-07
Output FRAME-RATE according to spec

daveisfera:
See https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.4.2


mauricioabreu:
Hello, @daveisfera thank you for contributing.

Reading the spec, it is not clear that it SHOULD be rounded to three decimal places. I understand the spec as saying it SHOULD not have more than three decimal places


https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices#Video
> 1.18. Video on Demand (VOD) content SHOULD use a natural frame rate for the content. Frame rates 23.976, 24, 25, 29.97, 30, 50, 59.94, and 60 fps are supported.

What is your opinion? cc @leandromoreira

daveisfera:
> Hello, @daveisfera thank you for contributing.
> 
> Reading the spec, it is not clear that it SHOULD be rounded to three decimal places. I understand the spec as saying it SHOULD not have more than three decimal places
> 
> https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices#Video
> 
> > 1.18. Video on Demand (VOD) content SHOULD use a natural frame rate for the content. Frame rates 23.976, 24, 25, 29.97, 30, 50, 59.94, and 60 fps are supported.

The Apple document is definitely worth considering but the [official spec reads](https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.4.2):

```
The value is a decimal-floating-point describing the maximum frame rate for all the video in the Variant Stream, rounded to three decimal places.
```

mauricioabreu:
@daveisfera I read the spec, but I didn't understand that as mandatory. 25 == 25,000

IMHO, Apple says that because in case the FPS has decimal places, it SHOULD have at most 3 decimal places.

daveisfera:
Ya, I can definitely see that. I was interpreting it to mean that there should always be 3 decimal places, but that's not what's actually stated there and the trailing 0s could be left off

mauricioabreu:
That's fine, @daveisfera. IMHO, specs are really hard to interpret.
I'm actually happy you took the time to open a pull request to solve a problem, even if it turned out not to be a problem at all.

Thank you for contributing to the m3u8 lib.

PR #362, opened 2024-05-09
Feat: add explicit 'Duration=' parameter to EXT-X-CUE-OUT

Equipo45:
# Description
I have been investigating the problem with DURATION parameters. 

Firstly, as you already know it isn't compulsory to establish the parameter as you can see in an example where is [use](https://api.encoding.com/reference/advanced-mbr-ad-breaks-manifests-hls) and other which [not](https://docs.aws.amazon.com/mediatailor/latest/ug/manifest-hls-example.html). The use of the parameter will depend on some key points:
1. Is there another parameter indicating two or more attributes about the ad break EX: `#EXT-X-CUE-OUT:DURATION=195.000,BREAKID=103038`
2. Does the parser or model used for generating the manifest support it?

I see the change as an improvement for the following reasons:
1. It **adds clarity to the parameter that is being introduced** by adding it explicitly.
2. Is currently used as a standard by [reference companies](https://cloud.google.com/video-stitcher/docs/concepts/ad-markers#ext-x-cue-out_and_ext-x-cue-in) such as Google.

Regarding all, thank you for making the package and contributing to the open-source community.

_A test has been changed for checking the new format_


m3u8/model.py
```diff
@@ -629,7 +629,7 @@ def dumps(self, last_segment, timespec="milliseconds"):
 
             output.append(
```
leandromoreira, line 630: I think we should preserve what the what we receive as input.

Meaning, we should expand [the parser to include something](https://github.com/globocom/m3u8/blob/master/m3u8/parser.py#L123) like `cue_out_duration_property` to true.

```python
        elif line.startswith(protocol.ext_x_cue_out):
            _parse_cueout(line, state)
            state["cue_out_start"] = True
            state["cue_out"] = True
            if line.contains(somethingsomething):
                   state["cue_out_explicitly_duration"] = True

```

And then here, we would need to check to see if we'd need to add the Duration or not. WDYT?

leandromoreira: Otherwise we'll transform manifests without Duration to have one.

leandromoreira (changes requested):
thanks, good work, I believe we must add the parser changes to this pr as well.

m3u8/model.py
Equipo45: Yeah, that's true. I changed it and adapted for parsing it only when DURATION is available. Also, a new test has been created and some code has been refactored regarding the new state.

leandromoreira (approved):
thanks for the great work

mauricioabreu (approved):
Awesome work, thank you!

PR #365, opened 2024-05-13
Add support for rendering REQ-VIDEO-LAYOUT in stream info classes

colekas-ssim:
REQ-VIDEO-LAYOUT is a relatively new tag within a stream info class - https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.4.3.8 (Page 48)


mauricioabreu (approved):
Thank you!

leandromoreira (commented):
Can you add the tag in the readme?

mauricioabreu:
@leandromoreira it is not a tag per se, it is a stream info attribute

colekas-ssim:
I can something to the README.rst, but it doesn't look like it has a section for these types of additions... open to suggestions.

mauricioabreu:
@colekas-ssim no need to. We currently just document tags, not specific attributes of each tag.

Thank you, merging!

PR #367, opened 2024-05-23
Add infspec for specifying precision for EXTINF

daveisfera:
At times, numerical quirks are output in `EXTINF`. I think that the `number_to_string` function should probably be made more conservative to handle this, but that's potentially a breaking change, so this PR adds `infspec` to allow specifying a desired precision of `milliseconds` or `microseconds` similar to `timespec`


mauricioabreu (approved):
Looks good, thank you for adding tests.

PR #368, opened 2024-06-13
Allow EXT-OATCLS-SCTE35 on non CUE-OUT segments

bbayles:
In https://github.com/globocom/m3u8/issues/85 and https://github.com/globocom/m3u8/pull/290, support for `EXT-OATCLS-SCTE35` tags was added - these are used by [Elemental](https://docs.aws.amazon.com/medialive/latest/ug/sample-manifests-hls.html#sample-manifests-hls-scte-35-enhanced) and some others.

Currently, this library only supports `EXT-OATCLS-SCTE35` tags on segments that in a "cue out" state. However, they are often used outside of that context as well - I've added an example playlist to the test suite.

This PR updates the support for these tags to allow them to flow through even if they aren't associated with a "cue out."


leandromoreira (approved):

m3u8/model.py
```diff
@@ -650,8 +649,10 @@ def dumps(self, last_segment, timespec="milliseconds", infspec="auto"):
             else:
                 cue_out_cont_suffix = ""
             output.append(f"#EXT-X-CUE-OUT-CONT{cue_out_cont_suffix}\n")
-        if self.cue_in:
+        elif self.cue_in:
             output.append("#EXT-X-CUE-IN\n")
+        elif self.oatcls_scte35:
+            output.append(f"{ext_oatcls_scte35}:{self.oatcls_scte35}\n")
```
mauricioabreu, line 655: Thank you for using f strings

mauricioabreu (approved):

bbayles:
Thanks! Looking forward to this being released.

mauricioabreu:
@bbayles I'm going to release a new version later today. Thank you!

PR #369, opened 2024-06-21
Refactor parser.py to use consistent arguments

bbayles:
This PR doesn't add any functionality, so I'll understand if you'd rather not take it. But I often found myself scrolling back and forth between sections of `parser.py` to look to see how a certain tag parser was called, and decided to try to make everything consistent.

The `parse` function now loops over the lines, checks for known tags, and always calls a parse function with these arguments:
* line
* lineno
* data
* state
* strict

The parse functions accept all of the arguments as keywords, and if they don't need them, they don't act on them. All state / data updating logic is handled inside the parse functions.

I find that this makes the `parse` loop easier to think about, and it makes the process of adding support for new tags easier.

### DIFF
--- a/m3u8/parser.py
+++ b/m3u8/parser.py
@@ -81,193 +81,173 @@ def parse(content, strict=False, custom_tags_parser=None):
         if len(found_errors) > 0:
             raise Exception(found_errors)
 
-    lineno = 0
-    for line in lines:
-        lineno += 1
+    for lineno, line in enumerate(lines, 1):
         line = line.strip()
+        parse_kwargs = {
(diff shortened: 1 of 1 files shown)


leandromoreira (approved):
thanks for all the hard work, i love it, maybe for the future we use a paramsObject instead of kwargs

mauricioabreu (approved):

mauricioabreu:
Amazing job @bbayles 
Could you rebase please? Then I can merge it immediately.

bbayles:
Should be one clean commit now - thanks!

Issue #371, opened 2024-07-25
'target_duration' returns float number, lead to iOS device fail to play the video


Scott0902:
The `target_duration` attribute in the `m3u8` library returns a float value, i.e.: 4.0, if I write a new playlist with tag `#EXT-X-TARGETDURATION: 4.0`, it will cause iOS devices fail to play the video. 

I discovered this bug by many attempts. Only by changing the `target_duration` to an integer resolves this issue, could allow iOS devices to play the video correctly.

I hope this m3u8 library to be modified to ensure the target_duration is always returned as an integer. This change will align the library with the HLS specification and ensure compatibility with iOS devices.

Thank you for your attention.

mauricioabreu:
Hello @Scott0902 could you send a playlist example so we can reproduce this behavior?

bbayles:
`EXT-X-TARGETDURATION` is indeed supposed to be an integer:
![image](https://github.com/user-attachments/assets/a15e866a-1db8-4f10-8635-99f92d379d0a)

The code currently calls [number_to_string](https://github.com/globocom/m3u8/blob/master/m3u8/model.py#L1650-L1658), which uses `to_integral_value` without a rounding specified. We could probably change it to use `ROUND_UP` for this particular value.

mauricioabreu:
@bbayles agreed! But I still wanna understand how `target_duration` was converted to `4.0`.

Here is the specific part where `target_duration` is converted to a number.

```python
if self.target_duration:
            output.append(
                "#EXT-X-TARGETDURATION:" + number_to_string(self.target_duration)
            )
```

```sh
>>> import m3u8
>>> m3u8.model.number_to_string("4")
'4'
>>> m3u8.model.number_to_string("4.0")
'4'
>>> m3u8.model.number_to_string("3.9")
'3.9'
```

Scott0902:
> Hello @Scott0902 could you send a playlist example so we can reproduce this behavior?

You can convert any H264-encoded `.mp4` video to `.ts` losslessly by using `ffmpeg.exe`, for example:

```
ffmpeg -i input.mp4 -c copy -bsf:v h264_mp4toannexb -f hls -hls_time 2 -hls_list_size 0 -hls_flags single_file -hls_segment_filename video.ts playlist.m3u8
```
By this way, `ffmpeg` will generate `playlist.m3u8` which is something like:

```
#EXTM3U
#EXT-X-VERSION:4
#EXT-X-TARGETDURATION:10
[...]

PR #372, opened 2024-07-29
Always parse target duration as integer

mauricioabreu:
Following the HLS spec, target duration is an integer value.


mauricioabreu:
Closes #371

m3u8/parser.py
```diff
@@ -447,7 +447,7 @@ def _parse_byterange(line, state, **kwargs):
 
 
 def _parse_targetduration(**parse_kwargs):
-    return _parse_simple_parameter(cast_to=float, **parse_kwargs)
```
bbayles, line 450: I don't think I understand how this rounds up rather than down?

mauricioabreu: I understood the problem does not happen when `dumps` get called. It actually happens when the playlist is parsed. `target_duration` was indeed being converted to float (as the diff shows)

bbayles: Oh, I see - in the `tests/playlists.py` change, you're updating the test data directly.

bbayles (approved):

m3u8/parser.py
mauricioabreu: Yes! Do you think that’s a bad approach? I changed the test playlist because, after reading the specifications, I believe a playlist should not have a target duration as a float.

bbayles: I think it's OK. At first I misunderstood the changes in this PR - I thought the test change was reflecting a new output behavior, but it was changed to reflect what's legal in the spec.

PR #374, opened 2024-07-30
Issue #373: Import EXT-X-CUE-OUT and EXT-X-CUE-OUT-CONT handling

### LINKED ISSUE #373: MediaConvert EXT-CUE-OUT-CONT still problematic
Hi,
I still have problems with some HLS manifests generated by AWS MediaConvert, even after the fix to #349.
Here is an example: https://bpkio-cs-demos.s3-eu-west-1.amazonaws.com/fabre/mediaconvert-conditioning/output/hls/cts-laworder720p.m3u8

The lib fails on this line:
`#EXT-X-CUE-OUT-CONT:10/4, SpliceType=VOD_DAI,Action=REPLACE,PAID=amazon.com/TEST2014020500000346,Acds=BA`.

Note that this type of markup is coming straight from the (poor) AWS documentation at https://docs.aws.amazon.com/mediaconvert/latest/ug/example-esam-xml-manifest-conditioning.html

bbayles:
This PR updates the handling for `EXT-X-CUE-OUT` and` EXT-X-CUE-OUT-CONT` tags.

Previously the parsers for these tags attempted to understand a few different vendor formats, but I've tried to simplify things to avoid crashing when novel ones are encountered.

Closes #373


m3u8/parser.py
```diff
@@ -320,9 +320,14 @@ def _parse_attribute_list(prefix, line, attribute_parser, default_parser=None):
 
     attributes = {}
     for param in params:
-        name, value = param.split("=", 1)
-        name = normalize_attribute(name)
+        param_parts = param.split("=", 1)
+        if len(param_parts) == 1:
```
bbayles, line 324: This change affects all tags - if there's some parameter without a `key=value` pair, it will get stuck into the empty string key.

tests/test_model.py
```diff
@@ -273,7 +281,7 @@ def test_segment_envivio_scte35_attribute():
 def test_segment_unknown_scte35_attribute():
     obj = m3u8.M3U8(playlists.CUE_OUT_INVALID_PLAYLIST)
     assert obj.segments[0].scte35 is None
-    assert obj.segments[0].scte35_duration is None
+    assert obj.segments[0].scte35_duration == "INVALID"
```
bbayles, line 284: We now no longer assume a particular format for duration, so weird fraction stuff can flow through.

mauricioabreu: That's great! Thank you

mauricioabreu (approved):

Issue #376, opened 2024-07-31
EXT-X-BITRATE parsing returns value as string rather than int


davemevans:
[EXT-X-BITRATE](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.4.4.8) is defined as a decimal integer describing the rate in kilobits per second.

Currently the parser will return this value [as a string](https://github.com/globocom/m3u8/blob/master/tests/test_parser.py#L859) which is awkward when working with this value.

I have a simple patch to parse and dump this as an int but it would technically be a small breaking change.

What do people think?

mauricioabreu:
@davemevans submit your PR, please. If it follows the spec, I'll gladly accept it. We can upgrade the major.

mauricioabreu:
Closed via #379

PR #379, opened 2024-08-07
Parse Bitrate value as integer not string

davemevans:
See #376.

Would break people who are doing `data["segments"][0]["bitrate"] == "1674"` (which is the intention), because we wish bitrate to be `int`.

Would also break people who are doing `obj.segments[0].bitrate = "9876"` because `dumps` explicitly requires `int` now, which I think is fine but could be changed to be more lenient.


mauricioabreu:
Thank you @davemevans nice fix.

Since it breaks users of m3u8, I'm going to release a major version again.

PR #382, opened 2024-09-24
Support Low Latency playlists with omitted LAST-MSN

bbayles:
I encountered [this stream](https://players.akamai.com/players/hlsjs?streamUrl=https%3A%2F%2Fstream.mux.com%2Fv69RSHhFelSm4701snP22dYz2jICy4E4FUyk02rW4gxRM.m3u8) in the wild. Its rendition reports are missing the `LAST-MSN` attribute, which causes an error with `m3u8`.

The spec [says](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.4.5.4):
```
The value is a decimal-integer specifying the Media Sequence
      Number of the last Media Segment currently in the specified
      Rendition.  If the Rendition contains Partial Segments then this
      value is the Media Sequence Number of the last Partial Segment.
      This attribute is REQUIRED.
```

But then it says:
```
   A server MAY omit adding an attribute to an EXT-X-RENDITION-REPORT
   tag - even a mandatory attribute - if its value is the same as that
   of the Rendition Report of the Media Playlist to which the EXT-X-
   RENDITION-REPORT tag is being added.  Doing so reduces the size of
   the Rendition Report.
```

Therefore it can be valid to leave it out, so `m3u8` is being too restrictive. I've fixed the issue in this PR.


mauricioabreu:
Awesome work @bbayles 
thank you and sorry for the late review.

PR #385, opened 2024-10-01
Adds support to python 3.13

jairhenrique:


daveisfera:
I didn't notice this already existed so I included it in #391

PR #386, opened 2024-10-08
Change m3u8 project to use poetry

DantasB:
- Change local installation of libraries, instead of that, using poetry.


bbayles:
For what it's worth, I've regretted doing this every time I've tried it.

DantasB:
lol, yeah, the idea of that is basically make things easier to external contributors, i mean, i'm not the one who loves installing a bunch of packages in my own environment as it in the future can cause some conflicts. 

Having a dedicated environment natively makes things easier and poetry helps that a little bit.

Issue #387, opened 2024-10-10
BUG since v3.5.0: Trying to load a playlist from file using an absolute path fails on Windows because of drive letters (URL-test in m3u8.load not implemented correctly!)


e-d-n-a:
I got an error while working with m3u8 v4.0.0 and trying to load a playlist from file using an absolute path on Windows.

### Traceback:
```
Traceback (most recent call last):
  File "[myscript.py]", line 843, in <module>
    asyncio.run(main())
  File "C:\Program Files\Python39\lib\asyncio\runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "C:\Program Files\Python39\lib\asyncio\base_events.py", line 647, in run_until_complete
    return future.result()
  File "[myscript.py]", line 806, in main
    await DLer.download_vod(file_pl, base_uri=url_pl_base, target=folder_target)
  File "[myscript.py]", line 478, in download_vod
    pl = m3u8.load(str(playlist), custom_tags_parser=self.__class__._parse_twitch_tags)
  File "C:\Program Files\Python39\lib\site-packages\m3u8\__init__.py", line 94, in load
    content, base_uri = http_client.download(uri, timeout, headers, verify_ssl)
  File "C:\Program Files\Python39\lib\site-packages\m3u8\httpclient.py", line 16, in download
    resource = opener.open(uri, timeout=timeout)
  File "C:\Program Files\Python39\lib\urllib\request.py", line 517, in open
    response = self._open(req, data)
  File "C:\Program Files\Python39\lib\urllib\request.py", line 539, in _open
    return self._call_chain(self.handle_open, 'unknown',
  File "C:\Program Files\Python39\lib\urllib\request.py", line 494, in _call_chain
    result = func(*args)
  File "C:\Program Files\Python39\lib\urllib\request.py", line 1417, in unknown_open
    raise URLError('unknown url type: %s' % type)
urllib.error.URLError: <urlopen error unknown url type: d>
```
... where `file_pl` was an absolute path to a m3u8-playlist on drive D (see 'd' in urlopen error). 
(same code worked, when`file_pl` was a relative path!)

I was dumbfounded to find out, that a bug got already introduced back in v3.5.0, that makes it impossible to load playlists from files (using absolute paths) on Windows, because the new method of distinction between URLs and file paths in m3u8.load just got changed for the worse.

[...]

PR #388, opened 2024-10-11
Issue 387: Support absolute Windows paths

bbayles:
Re: [Issue 387](https://github.com/globocom/m3u8/issues/387), this PR changes the "is this a URL?" test to look for another required element: the `netloc`. This fixes the issue because absolute Windows paths will not have such an element.

I've added a test to ensure that this doesn't regress in the future.


mauricioabreu:
Thank you for another amazing contribution!

tests/test_loader.py
```diff
@@ -146,3 +149,14 @@ def test_raise_timeout_exception_if_timeout_happens_when_loading_from_uri():
         assert True
     else:
         assert False
+
+
+def test_windows_paths():
+    file_path = "C:\\HLS Video\test.m3ui8"
```
e-d-n-a, line 155: I think, this should be:
```file_path = "C:\\HLS Video\\test.m3u8"```

lee-cq:
Can this update be published on PyPI?

PR #391, opened 2024-11-08
EOL 3.8

daveisfera:


mauricioabreu:
so versions lower than 3.9 are no longer supported. Right? I'm lost in time 😄

daveisfera:
Yes, 3.8 was EOL last month
https://devguide.python.org/versions/

mauricioabreu:
Thank you @jairhenrique and @daveisfera

PR #393, opened 2024-12-17
Add dependabot

daveisfera:


daveisfera:
Similar to #384 but this one also runs `dependabot` on `pip`

PR #394, opened 2024-12-17
Use pip-tools to lock down versions

daveisfera:
Similar to #386 but using `pip-tools` (and will allow easier version management when combined with #393 )


daveisfera:
One of the tests failed, but I think it was a transient error, because the versions installed are exactly the same as the tests that ran for #393 and it passed without issue

PR #399, opened 2025-10-03
feat: add EXT-X-DEFINE implementation

SirMonteiro:
As defined on https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.4.2.3.
I saw that globocom/go-m3u8 implements this but python one was missing!


SirMonteiro:
Please, add hacktoberfest-accepted label for this PR, or add hacktoberfest label for the repo.
I really want to contribute, but at same time, participate of the event!

